PR #7, opened 2021-05-01
Map interface and performance improvements.

StarArawn:
For the 0.3 release. New stuff:
- Improved rendering by meshing chunks in the vertex shader.
- Chunk vertices are now calculated across threads instead of 1 thread.
- GPU Animations
- Chunk Camera Culling
- Tile visibility
- More examples 
- Better docs
- Cargo fmt code.


src/map.rs
```diff
@@ -14,29 +14,47 @@ pub struct MapBundle {
...
     pub texture_size: Vec2,
     pub layer_id: u32,
-    chunks: HashMap<MapVec2, (Entity, Vec<Option<Entity>>)>,
     pub mesher: Box<dyn TilemapChunkMesher>,
+}
+
+impl Clone for MapSettings {
+    fn clone(&self) -> Self {
+        Self {
+            size: self.size.clone(),
```
TheLeonsver1, line 29: I think these `Vec2`s should be `Copy` instead of `.clone()` the compiler would optimize this on release, but if I'm not mistaken, wouldn't optimize it on debug.

also, is it possible to rename `size` to `map_size`?

src/chunk.rs
```diff
@@ -202,3 +204,65 @@ pub(crate) fn update_chunk_mesh(
...
+    if let Some((_current_camera, ortho, camera_transform)) = camera.iter().find(|data|
+        if let Some(name) = &data.0.name {
+            name == CAMERA_2D
+        } else { false }
+    ) {
+        // Transform camera into world space.
+        let left = camera_transform.translation.x + (ortho.left * camera_transform.scale.x);
+        let right = camera_transform.translation.x + (ortho.right * camera_transform.scale.x);
+        let bottom = camera_transform.translation.y + (ortho.bottom * camera_transform.scale.x);
+        let top = camera_transform.translation.y + (ortho.top * camera_transform.scale.x);
```
TheLeonsver1, line 221: I think this is an error, you have `camera_transform.scale.x` even for `top` and `bottom`, and I mean I don't think the scale would be different on the x and y but for being safe this should probably be changed to `camera_transform.scale.y` in the top and bottom parts

PR #10, opened 2021-05-02
Added a MapSettings struct for convenience

### LINKED ISSUE #8: Map settings reusability improvement suggestion
Currently `Map` isn't `Clone` (and I don't think it should be, or at least not for this issue's purposes) 
and `Map::new()` currently takes:
```
size,
chunk_size,
tile_size,
texture_size,
(layer_id isn't relevant to this issue)
```

My suggestion would be to create a `MapSettings` struct that is `Clone`/`Copy`:
```rust
///The settings for tilemap layers
#[derive(Debug, Clone, Copy)]
pub struct MapSettings {
    pub map_size: MapVec2,
    pub chunk_size: MapVec2,
    pub tile_size: Vec2,
    pub texture_size: Vec2,
}
```
To enable making similar maps with the same settings more convenient(for example if you need layers with the same settings). The current impl kind of forces the user to create functions to reduce boilerplate which I think should be avoided)

This also enables storing these structs as/in resources for reusability
`Map::new()` would then take `MapSettings` instead of the 4 different values

*I don't mean you need to actually store this inside of `Map` instead of the current configuration. 
you can and maybe should keep `Map` as is and only change `Map::new()` to make it more convenient for users. I'm not sure.. it might be nice to copy the settings of a `Map` if it's possible, idk if there'd be uses for that, maybe for `bevy_tiled` maps..?

TheLeonsver1:
closes #8

### DIFF
--- a/examples/accessing_tiles.rs
+++ b/examples/accessing_tiles.rs
@@ -13,10 +13,21 @@ fn startup(
 
     let texture_handle = asset_server.load("tiles.png");
     let material_handle = materials.add(ColorMaterial::texture(texture_handle));
-
-    let mut map = Map::new(Vec2::new(2.0, 2.0).into(), Vec2::new(9.0, 9.0).into(), Vec2::new(16.0, 16.0), Vec2::new(96.0, 256.0), 0);
+    let map_settings = MapSettings {
+        map_size: Vec2::new(2.0, 2.0).into(),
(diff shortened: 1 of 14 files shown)


TheLeonsver1:
Closed since it's done in #7 (3ac42b471264fe62b5359fc1252b91013e22ecb2)

PR #11, opened 2021-05-02
Add "web" feature for wasm support

rparrett:
There are two main things going on here

- swapping in a separate set of opengles compatible shaders
- using default-features=false for bevy (`bevy_wgpu` and perhaps others won't compile for wasm32)

I think that it should be possible to clean this up using cargo's `resolver="2"` new feature resolver. I think It **should** be possible to:

- use e.g. `[target.'cfg(target_arch = "wasm32")'.dependencies]` instead of adding a web feature
- make `bevy_wgpu`, `bevy_winit`, and `bevy_png` dev-dependencies

But I couldn't manage to make it work, and ran out of time.

Also, I only tested with one of the examples slapped into [bevy_webgl2_app_template](https://github.com/mrk-its/bevy_webgl2_app_template)

Please feel free to close / edit / whatever.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,22 @@ exclude = [
     "screenshots/*",
 ]
 
+[features]
+
+default = [
+    "bevy/bevy_wgpu",
+    "bevy/bevy_winit",
+    "bevy/png",
+    "bevy/render",
+]
+
+web = [
+    "bevy/bevy_winit",
+    "bevy/render",
+]
+
 [dependencies]
-bevy = "0.5"
+bevy = { version = "0.5", default-features = false }
 log = "0.4"
 dyn-clone = "1.0"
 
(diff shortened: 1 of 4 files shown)


StarArawn:
This looks pretty good my only concern is that there isn't a way of testing this code? I don't really want to shove a web example in this code base either, but perhaps I should create an example in a separate repo. The `feature resolver` sounds useful as well, hopefully we can figure that out.

rparrett:
Yeah, I share that concern but I'm not immediately sure what to do about it.

martinfiers:
Just fwiw, I was able to get it working (based on the latest main), by simply using the following pattern (it didn't require any changes to the Cargo.toml or the introduction of a new feature, and works well when compiling vs wasm or native):

```rust
#[cfg(not(target_arch = "wasm32"))]
create_chunk_pipeline!(
    SQUARE_PIPELINE,
    8094008129742001941,
    create_square_pipeline,
    "square-tilemap.vert",
    "square-tilemap.frag"
);

#[cfg(target_arch = "wasm32")]
create_chunk_pipeline!(
    SQUARE_PIPELINE,
    8094008129742001941,
    create_square_pipeline,
    "square-tilemap.es.vert",
    "square-tilemap.es.frag"
);

```

The vertex/fragment shaders are pretty easy to port (since they became a bit more complex recently), one detail is that I had to rename gl_VertexIndex to gl_VertexID (they seem to be equivalent) and some float conversion.

I don't really have the time to make a PR but I thought this information might be useful.

rparrett:
I'm going to close this for now, as I'm lacking the time/motivation to update it currently.

Issue #19, opened 2021-05-11
Feature: Isometric Staggered


martinmolin:
It would be really useful to be able to select Isometric Staggered or Isometric Diamond (like the current iso example) when creating a map.

For reference:
[Isometric tilemap types](https://gamedev.stackexchange.com/questions/49847/difference-between-staggered-isometric-and-normal-isometric-tilemaps)

StarArawn:
I'm trying to wrap up some bugfixes so I don't have the time currently to add this, but I can at least point to where things are happening if you are up for making the change yourself. For example diamond iso happens here:
https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/src/render/iso-tilemap.vert#L32

You would need to duplicate the `iso-tilemap.vert` shader and call it something like: `iso-staggered-tilemap.vert`. Make the changes to the position so it outputs staggered instead of diamond. Finally create a new pipeline using the new shader and match the `IsoType` to the correct pipelines. I should be able to add this feature perhaps this weekend, but I'm not super familiar with staggered positioning of iso tiles so I would need to try and find some resources for how that is suppose to work. It kinda looks very similar to hex odd/even row positioning though.

martinmolin:
Yep! Like you said, instead of doing:
```
position.xy *= tile_size;
```
In the shader it seems to work if you do:
```
position.x *= tile_size.x;
position.y *= tile_size.y / 2.0;
```
![bild](https://user-images.githubusercontent.com/670540/117869248-dc8f0400-b29a-11eb-85f9-3f983a32d689.png)

I can try to fork and do these changes there and make a PR that you can take a look at whenever you have time!
Thanks for the pointers 👍

StarArawn:
The tiles look a little skewed now. I think the positioning is likely correct, but you'll want to still make sure the tiles are still scaled to `tile_size`, and only the positioning of the tile is changed.

martinmolin:
Yep you're right! Good catch 😄

martinmolin:
After a bit of fiddling I ended up with:
![bild](https://user-images.githubusercontent.com/670540/117877996-8d020580-b2a5-11eb-93af-66c6cfee6948.png)

I'll make a PR and you can check it whenever!

ekaats:
[...]

PR #20, opened 2021-05-11
Feature: Isometric Staggered

martinmolin:
This PR is an implementation for #19 .

I tried to use the same naming conventions as the different hex related names.
One thing I'm not quite sure is how you get a pipeline_id, I just typed a random number with the same length as the others.
Another thing is if the positioning is good overall. It's hard to tell when not using a map that you know how it should look!

- Renamed the iso example to iso_diamond
- Created a new iso_staggered example
- Added IsoType (similar to HexType)
- Used the hex row odd vertex shader as a base, only slight tweaks for y offset needed for a staggered iso map

iso_staggered example running:
bild


StarArawn (approved):
Looks good to me thank you for the work to get this done! 👍

Issue #23, opened 2021-05-13
Add height tiles to isometric example


ekaats:
The isometric example only allows for flat tiles at the moment. I have a fix ready for the renderer but am still working on placing the correct tiles. For this I have to look into how the tile atlas works so I can place the correct adjacent tiles. This may make the example a lot more complicated.

demilich1:
Cannot comment on the linked fix, but I just wanted to say that I have just encountered this issue in my project. I think this is a requirement for most isometric games, for now I have to abandon bevy_ecs_tilemap unfortunately :( Also related to #44

StarArawn:
This should be fixed now. There are some bugs with Iso tilemaps that have height but they are being worked out.

PR #24, opened 2021-05-13
Isometric tilemap height correction

ekaats:
I've made some changes to the isometric rendering to allow for tiles that are not flat to be rendered correctly. In the example I've added some random foliage to show the effect. Still need to figure out the tilemap coordination system, so I was not able to create nice hill/mountain ranges with the correct tiles connecting to each other.

I'm still very new to rust so it is very likely that my approach is not the best.

### DIFF
--- a/examples/iso.rs
+++ b/examples/iso.rs
@@ -10,35 +10,51 @@ fn startup(
     mut meshes: ResMut<Assets<Mesh>>,
     mut materials: ResMut<Assets<ColorMaterial>>,
 ) {
+    // Create the camera setup
     commands.spawn_bundle(OrthographicCameraBundle::new_2d());
 
+    // Load textures
     let texture_handle = asset_server.load("iso.png");
+    
+    // Apply textures to materials
     let material_handle = materials.add(ColorMaterial::texture(texture_handle));
 
+    // Set a ratio for texture size. Higher means chunkier textures
+
+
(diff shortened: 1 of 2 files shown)


ekaats:
This closes #23

StarArawn:
I've been thinking about this for a while, and I'm not yet convinced this is the correct approach.. The iso tilemap size for the ground here is "suppose" to be 64x32, I think for the the "walls" however it might make more sense to have 64x64 tiles. 🤔

ekaats:
That is a fair point. I do wonder whether they actually use different sizes for floor and wall tiles in that tutorial though.  I quickly read through it and it appears they use the same but point out you'd need to draw different tiles compared to the top-down perspective. For example, they say the following: "(Here, we essentially assume that tile height and tile width are equal, as in most cases.) "

The nice thing about loading the tiles as the same height is that you do not need to worry about the difference between the two types of tiles:
bevy_height

Only bigger sprites would need to be handled differently.

StarArawn:
I was looking at how some other isometric 2D games handle rendering and I think I want to rethink and maybe provide some higher level support for iso. Typically draw order is very important as you want to have "walls" potentially draw on-top of sprites. I think we need some mechanisms for translating a 2D point to a 3D point for iso. Also we should likely render each layer from the top down in "sections" at least for diamond. I'm not sure how you would even do that with my current implementation.

Issue #18, opened 2021-05-11
Tile entities are leaked upon recursively despawning the Map entity


forbjok:
When the Map entity is despawned with .despawn_recursive(), all tile entities belonging to that tilemap continue to exist even though they are no longer visible. As far as I can tell, this is due to the tile entities not being marked as children of the Map entity.

forbjok:
They don't necessarily need to be children of chunks, do they?
Both chunks and tiles could all just be children of the Map entity itself.

StarArawn:
Part of the issue with having tiles be a child of the map entity or the chunk entity is that tiles don't have standard transforms, and bevy currently complains about that. We could give them standard transforms, but then its just a lot of extra data that is not needed and people might assume that changing a tile transform would change where a tile is located which isn't supported and isn't easy to support.

forbjok:
> tiles don't have standard transforms, and bevy currently complains about that

If that's the case, maybe we should suggest fixing it in Bevy. I don't see any reason why entities should be forced to have a Transform in order to be part of a hierarchy.

StarArawn:
> > tiles don't have standard transforms, and bevy currently complains about that
> 
> If that's the case, maybe we should suggest fixing it in Bevy. I don't see any reason why entities should be forced to have a Transform in order to be part of a hierarchy.
Bevy relations RFC would fix this issue. :)

See:
https://github.com/bevyengine/rfcs/pull/18

StarArawn:
@forbjok I ran some tests with this, and I discovered a performance drop almost in half due to having a parent per child. If I instead have a `map_query.despawn_layer(layer_id)` would that be acceptable?

forbjok:
Interesting. I wonder why that would affect performance when tiles don't have transforms that would need to be propagated. Have you looked into why it's affecting performance?

StarArawn:
The performance is mainly lost due to how `bevy_transform` propagates the transform information for parents/children. It always iterates over the children regardless of whether they have transforms or not.

StarArawn:
You can now use `map_query.depsawn_map` to despawn an entire map including tiles. Optionally you can despawn layers with `depsawn_layer` or tiles with: `despawn_layer_tiles`.

PR #26, opened 2021-05-13
Fixed tile entities leaking when map entity is despawned recursively

forbjok:
I did some testing regarding https://github.com/StarArawn/bevy_ecs_tilemap/issues/18, and as far as I can tell Bevy does not have a problem with having a Parent component on entities without a Transform.

This pull request makes .add_tile() insert a Parent component on all tile entities so that they are automatically despawned when the map entity is despawned recursively. It does not insert a Parent component on chunks, but at least in my use case, this seems to completely fix the issue of orphaned tile entities being left behind.

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -95,6 +95,7 @@ impl Clone for MapSettings {
 pub struct Map {
     /// The map information for the tilemap entity.
     pub settings: MapSettings,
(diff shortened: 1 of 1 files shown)


StarArawn:
Thank you, for the work on this! it looks pretty solid! 👍

PR #28, opened 2021-05-19
Api interface changes

StarArawn:
WIP


TheLeonsver1:
I think we should split `MapQuery` into two `SystemParam`s for better parallelization of systems that run about tiles, one that can mutate the layers and such, and one that lets us only read the tilemap information, since queries on `&mut Chunk` can not run in parallel since they target the same entities.

src/chunk.rs
```diff
@@ -140,7 +140,16 @@ impl Chunk {
         mesher: Box<dyn TilemapChunkMesher>,
         cull: bool,
     ) -> Self {
-        let tiles = vec![None; chunk_size.x as usize * chunk_size.y as usize];
+        let tile_size_x = (1
```
TheLeonsver1, line 143: I think this should be put inside a static function outside, since I think it is duplicated twice in the project

src/layer_builder.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub fn get_tile_entity(&mut self, tile_pos: UVec2) -> Result<Entity, MapTileError> {
+        let morton_tile_index = morton_index(tile_pos);
+        if morton_tile_index < self.tiles.capacity() {
+            return Ok(self.tiles[morton_tile_index].0);
+        }
+
+        Err(MapTileError::OutOfBounds)
+    }
+
+    /// Gets a reference to the tile data using the a tile position.
```
TheLeonsver1, line 67: There's a type here, should probably be "Gets a reference to the tile data using the tile position."

examples/accessing_tiles.rs
```diff
@@ -20,63 +20,95 @@ fn startup(
...
+    let layer_entity = commands.spawn().id();
+
+    // Create the layer builder
+    let mut layer_builder = LayerBuilder::<TileBundle>::new(
         &mut commands,
-        &mut meshes,
-        material_handle,
-        map_entity,
+        layer_entity,
+        LayerSettings::new(
```
TheLeonsver1, line 38: I think that `LayerBuilder::new()` should take `layer_id` as a parameter, and we also should probably move `layer_id` out of `LayerSettings` since it's not really data that can/should be reusable between layers.

src/layer_builder.rs
```diff
@@ -0,0 +1,278 @@
...
+    pub(crate) tiles: Vec<(Entity, Option<(T, bool)>)>,
+    pub(crate) layer_entity: Entity,
+}
+
+impl<T> LayerBuilder<T>
+where
+    T: TileBundleTrait,
+{
+    /// Creates the layer builder using the layer settings.
+    pub fn new(commands: &mut Commands, layer_entity: Entity, settings: LayerSettings) -> Self {
```
TheLeonsver1, line 22: I think that LayerBuilder::new() should take layer_id as a parameter, and we also should probably move layer_id out of LayerSettings since it's not really data that can/should be reusable between layers.

StarArawn: I do want people to still be allowed to call build themselves directly on the builder. `map_query.create_layer` is just a nice to have convivence feature.

TheLeonsver1: yeah, I didn't notice `build()` yet, then yeah, we should probably add the `layer_id` as a parameter in `LayerBuilder::new()` then probably.

src/layer_builder.rs
```diff
@@ -0,0 +1,278 @@
...
+            if let Some(tile) = &self.tiles[morton_tile_index].1 {
+                return Ok(&tile.0);
+            } else {
+                return Err(MapTileError::NonExistent);
+            }
+        }
+        Err(MapTileError::OutOfBounds)
+    }
+
+    fn get_tile_i(&self, tile_pos: IVec2) -> Option<(bevy::prelude::Entity, &T)> {
```
TheLeonsver1, line 80: I think it would be nicer to name this `get_tile_i32()` (if we want to stick with glam's naming style, like `as_i32()` for example) or get_tile_ivec(), `i` is always used to name indexes so the function name is a bit confusing..

TheLeonsver1:
I think we should consider adding to `LayerBuilder` functions like `new_with_set_all()` and `new_with_fill()`.
The current combination of `new()` and then `set_all()` is not the most efficient on bigger maps since we need to iterate over every tile twice,
This also might mean more lookups for the ECS behind the scenes because we might hurt the cache locality of applying commands on our entities? (not knowledgeable enough about the ECS.. sorry..).

At the same time, I think filling(as in `LayerBuilder::fill()`) the map in the same iteration as spawning would be faster too.

TheLeonsver1:
I think we should add functions like 'new_with_set_all()` and `new_with_fill()` to `LayerBuilder` to have less iterations on tiles,
we are already iterating on them twice with `new()` and `build()` so it's a bit inefficient and can make initiating bigger maps slower on older CPUs.

src/layer_builder.rs
TheLeonsver1: We should make sure to add `layer_id` to `new_and_build()` too :)

TheLeonsver1:
We've decided to try and see if giving the user the `chunk_entity`, and with that letting them get away with using only one `insert_bundle()`  would help with performance by perhaps having fewer `archetype` changes per tile entity

PR #38, opened 2021-05-30
Don't include default features when used as a dependency.

ndarilek:


StarArawn:
This looks good, but can you update the description to explain why this is useful. I think I know why, but I think it would be good to have documented here. 👍

ndarilek:
I thought I covered that in the commit description. Do you need 
something in addition to that?


Essentially, your line in Cargo.toml depends on Bevy with all features 
enabled, E.g. audio. I've stripped the audio feature from my own game 
since I roll my own, but depending on any plugin that just depends on 
bevy=0.5 compiles it anyway even if it isn't used.


Thanks.

PR #49, opened 2021-06-09
Fix panic on attempting to fetch out of bound tiles.

### LINKED ISSUE #48: MapQuery's get_tile_entity crashes with an index out of bounds instead of returning MapTileError::OutOfBounds
When you attempt to call get_tile_entity with an index that is out of bounds, it panics instead of returning MapTileError. (eg. on a 12x12 tile map a tile_position of (17, 0) will cause it to panic. (16, 0) will also panic, but (15, 0) won't panic).

Elabajaba:
Fixes #48 

The * and ? were to avoid having a nested option, and to change the least amount of code possible.
I've only tested it on get_tile_entity, but it seems to work.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -127,7 +127,10 @@ impl Layer {
     }
 
     pub fn get_chunk(&self, chunk_pos: ChunkPos) -> Option<Entity> {
-        *self.chunks.get(morton_index(chunk_pos))?
+        match self.chunks.get(morton_index(chunk_pos)) {
+            Some(Some(chunk)) => Some(*chunk),
+            _ => None,
+        }
     }
 
     /// Gets the map's size in tiles just for convenience.
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -127,7 +127,7 @@ impl Layer {
     }
 
     pub fn get_chunk(&self, chunk_pos: ChunkPos) -> Option<Entity> {
-        self.chunks[morton_index(chunk_pos)]
+        *self.chunks.get(morton_index(chunk_pos))?
     }
 
     /// Gets the map's size in tiles just for convenience.


src/layer.rs
```diff
@@ -117,7 +117,7 @@ impl Layer {
     }
 
     pub fn get_chunk(&self, chunk_pos: UVec2) -> Option<Entity> {
-        self.chunks[morton_index(chunk_pos)]
+        *self.chunks.get(morton_index(chunk_pos))?
```
StarArawn, line 120: I'm not sure this is a great fix. I need to think about it a bit more. 🤔

PatrickLerner:
Would love to see it fixed in next release as it breaks the 'get neighbors' function for tiles on the border. This limits the usability of this function in many algorithms (e.g. game related systems that use neighbors or things like pathfinding) if you have to handle border edges differently.

The only way to work around it for me was to keep the map sparse and not assign the outermost tiles and ignore them completely.

src/layer.rs
anhyzer2602: I ran into this as well while playing around with things. Does it make more sense to return either a Result<Entity, MapTileError> rather than Option<Entity>? This way OutOfBounds can be returned for this particular scenario and NonExistent could be returned in place of a None value.

Elabajaba: Would this be acceptable?

```rust
    pub fn get_chunk(&self, chunk_pos: ChunkPos) -> Option<Entity> {
        match self.chunks.get(morton_index(chunk_pos)) {
            Some(Some(chunk)) => Some(*chunk),
            _ => None,
        }
    }
```

StarArawn: I'm planning on ripping out the morton encoding(I think it slows things down overall), but I think this a viable solution.

Elabajaba:
I updated this to use a match instead so it's clearer. Is there any chance this could get merged in case bevy 0.6 releases before you rip out morton encoding?

PR #51, opened 2021-06-19
Wasm support

ypoonse:
Hello,

A (second) attempt at supporting wasm. Some inspiration on the shaders thanks to #11  ( @rparrett ).

I wasn't sure if you still prefer an example in a separate repo or not. If you do, then I can remove it again. The benefit now though is that it's a bit easier to test during development. The example_webgl2 folder is created based on https://github.com/mrk-its/bevy_webgl2_app_template. 

Currently I only ported the isometric grid vertex and fragment shaders - mainly because those are the ones I'm using in my project and I'm testing them regularly. It's mainly the testing of the others on wasm that's a bit cumbersome - they could be added later on-demand? Let me know what you think.


StarArawn (commented):
What happens for all of the other shaders? Does webgl throw some kind of error for non-square tilemaps?

ypoonse:
I just tried this, I then get this error in the browser console:

```
 C:\Users\mfiers\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_webgl2-0.5.2\src\renderer\webgl2_render_resource_context.rs:76 compiling shader: "#version 450\r\n#define COLORMATERIAL_TEXTURE\n#define WEBGL [...]" snomblib.js:369:17

panicked at 'called `Result::unwrap()` on an `Err` value: "ERROR: 0:1: \'\n\' : invalid version directive\nERROR: 0:6: \'layout\' : syntax error\n"', C:\Users\mfiers\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_webgl2-0.5.2\src\renderer\webgl2_render_resource_context.rs:77:81

```

The error is quite logical, but of course not really super-helpful. It's not really possible to check at compile-time as when you load a Tiled file you don't know upfront what tile type it will be.

ypoonse:
@StarArawn any suggestions on how you want to move this forward? I don't really have any suggestions on how to improve testing - except I can say I would use this for my own project :). Is it required for you to implement all tile types?

PR #59, opened 2021-06-28
Adds simple readme badges

actioninja:
Adds some of those nifty badges like quite a few other bevy crates use to the readme.

Provides a bit easier of access to the Crates.io page and docs.rs for the plugin.


StarArawn (approved):
Thanks these look great! 👍

Issue #14, opened 2021-05-06
API enhancement: add MapSettings::new_with_map_type()


TheLeonsver1:
Currently `MapSettings::new()` is locked to the `TilemapMeshType::Square` map type.
It would be nice to have a `MapSettings::new_with_map_type()` that would accept the different `TilemapMeshType` options.

StarArawn:
No longer relevant with the new API.

PR #71, opened 2021-07-18
#14 LayerSettings::new_with_map_type

martinmolin:
Implementation for #14 , should this be named `new_with_mesh_type` or is `new_with_map_type` better?

- Adjusted hex and iso examples to use the new function.
- Corrected window titles for the iso_staggered and hex_row examples.

### DIFF
--- a/examples/hex_column.rs
+++ b/examples/hex_column.rs
@@ -19,13 +19,13 @@ fn startup(
     let map_entity = commands.spawn().id();
     let mut map = Map::new(0u16, map_entity);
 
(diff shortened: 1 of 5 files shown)


PatrickLerner:
Maybe it would be nice to think about a builder-pattern here? It's already a bit cryptic what all those `UVec2` mean without looking at the docs or using some editor support to give you the parameter names?

```
        UVec2::new(2, 2),
        UVec2::new(64, 64),
        Vec2::new(17.0, 15.0),
        Vec2::new(102.0, 15.0),
```

PR #81, opened 2021-08-21
Updated MapQuery to fit with bevy main's SystemParam's split lifetimes change

TheLeonsver1:
Should I anchor this pr to the last rev?


StarArawn:
This looks good. Can we make it a full PR? I think having it how it is makes sense as I'll likely merge this first and then merge main into `bevy-track`.

TheLeonsver1:
What do you mean by making it a full pr? undrafting it? if so, i undrafted :D

StarArawn (approved):

PR #90, opened 2021-08-24
Add borders to tiles

### LINKED ISSUE #86: Add borders to default assets
This will make it easier for users to clearly understand the effects of changes to their code when experimenting with the examples.

### LINKED ISSUE #88: Texture size in map.rs example is wrong
Texture size is actually 96 x 16, but is listed as 96 x 256.

This silently works for the monochrome tiles, but changes made do not show up properly.

See also: #87 ;)

alice-i-cecile:
Fixes #86.
Also fixes #88, as the tiles are incorrectly rendered without the fix.


alice-i-cecile:
Look at the increased clarity! 
image

Still ugly, but much more useful for debugging and understanding.

StarArawn (approved):

PR #93, opened 2021-08-26
Add traits for map and layer IDs

alice-i-cecile:
These traits are now used instead of `Into<u16>` in the end-user-facing APIs. This clarifies intent over just a `Into<u16>`, allows engine and downstream uses to perform more logic on map and layer ids (as more trait bounds are used), and prevents users from accidentally swapping a map and layer identifier.

These traits are implemented for `u16` by the library itself to avoid breaking user code unnecessarily. I didn't change the example code to use enums as I felt that was out of scope for this PR.

The other design change you could make on top of this is enforcing / encouraging a unified map / layer ID schema by adding map and layer identifier generic types to `TilemapPlugin`. These would be passed in by users when declaring the plugin, and would help avoid clashes.

I don't think this is terribly likely to occur in practice, but it's worth briefly discussing as it communicates the intended usage a bit more clearly IMO. We could also possibly migrate away from using `u16` internally for this at all by taking this approach.

### DIFF
--- a/src/map_query.rs
+++ b/src/map_query.rs
@@ -63,13 +63,13 @@ impl<'a> MapQuery<'a> {
     ///   tile.texture_index = 10;
     /// }
     /// ```
-    pub fn set_tile<M: Into<u16> + Copy, L: Into<u16> + Copy>(
+    pub fn set_tile(
         &mut self,
         commands: &mut Commands,
         tile_pos: TilePos,
         tile: Tile,
-        map_id: M,
-        layer_id: L,
+        map_id: impl MapId,
+        layer_id: impl LayerId,
     ) -> Result<Entity, MapTileError> {
         let map_id = map_id.into();
         let layer_id = layer_id.into();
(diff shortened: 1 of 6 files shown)


StarArawn:
I think if you just merge `Main` into this it should be good to merge! 👍

alice-i-cecile:
Rebased! Should be good to go.

src/map_query.rs
```diff
@@ -328,17 +328,91 @@ impl<'a> MapQuery<'a> {
...
+    ///
+    /// ## Example
+    ///
+    /// ```
+    /// let neighbors = map.get_tile_neighbors(UVec2::new(0, 0));
+    /// assert!(neighbors[1].1.is_none()); // Outside of tile bounds.
+    /// assert!(neighbors[0].1.is_none()); // Entity returned inside bounds.
+    /// ```
+    pub fn get_tile_neighbors<M: Into<u16>, L: Into<u16>>(
+        &self,
```
StarArawn, line 346: This looks like a bad merge/rebase to me. Didn't this code move into a separate file?

StarArawn (changes requested):

src/map_query.rs
alice-i-cecile: Yep. I'll go over it and check again.

alice-i-cecile:
Fixed the merge and removed the duplicated code! It compiles now, and the examples I checked (Game of Life, the basic map examples) run as expected.

PR #96, opened 2021-09-09
Convert TilePos, MapSize and relatives to use usize

### LINKED ISSUE #95: TilePos and relatives should use usize, not u32
This is an important perf and usability optimization when interfacing with external arrays (which are always indexed by `usize` slices).

Currently, you must repeatedly waste operations performing this conversion and littering the code base with `as usize`.

alice-i-cecile:
Fixes #95.

This will make the end-user API dramatically nicer (and hopefully faster) when interfacing with resources that want to be accessed by `TilePos` coordinates.

Incidentally, this also expands the maximum size of the map to `u64` in most cases. This is almost certainly a *bad* idea for users to take advantage of though, as 2^32 tiles is already monstrous and Bevy will run out of `Entity` IDs to take advantage of.

There are two main drawbacks here:
  - doubles the memory footprint of coordinates. `MapSize` etc are irrelevant here, but `TilePos` is not
  - forces a more expensive conversion to `UVec2` API. It's not immediately clear to me that we need `UVec2` at all in the stack, but I didn't dig deep enough to investigate. Without a solid set of benchmarks to run, I'm reluctant to try to make perf-driven changes like this.
  
If these drawbacks are too severe; we could work around this in end user code by doing the conversions there, probably with a decent API.  This incurs a different set of reasonably serious perf trade-offs, but will only be relevant in some end user applications.

### DIFF
--- a/src/layer.rs
+++ b/src/layer.rs
@@ -1,5 +1,6 @@
 use crate::{
     chunk::Chunk,
+    map::MapId,
     morton_index,
     prelude::{ChunkMesher, Tile},
     round_to_power_of_two,
@@ -7,6 +8,7 @@ use crate::{
     ChunkPos, ChunkSize, MapSize, TextureSize, TilePos, TileSize, TilemapMeshType,
 };
 use bevy::prelude::*;
+use std::hash::Hash;
 
 /// A bevy bundle which contains: Map, Transform, and GlobalTransform components.
 #[derive(Bundle, Default)]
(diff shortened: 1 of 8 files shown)


alice-i-cecile:
I think I want to hold off merging this until we can pull together a basic benchmark or three to see what sort of run-time impact the UVec2 conversion step has. If it's bad, I'll see if I can fix it.

An extra 8 bytes per tile entity isn't much memory overhead on the scale of things. At a million entities, that's about 8 MB; I think you'd have to hit about 100 million before I cared about the added memory footprint as an end user at all. I would expect that scale to be exceedingly rare, and even then the extra CPU overhead from having to do conversions when slicing into a resource constantly is probably a lot worse.

alice-i-cecile:
Closing for now. I still think it's a good idea, but I don't have time.

PR #102, opened 2021-10-11
fix: texture bounds on atlas with spacing

icorderi:
When an atlas texture had spacing the column calculation and sprite offsets were not being calculated correctly.

This patch fixes the math required to correctly calculate the texture bounds for the tile in the atlas on Square tile maps.


StarArawn:
Can we add an example so we can properly test this in the future? I think just adding a new example with a new texture that has 1 pixel of spacing around the existing square tiles should be enough.

PR #104, opened 2021-10-11
Move Tiled and LDTK functionality to examples

### LINKED ISSUE #84: Remove editor specific code in favor of including them as basic examples.
I've had a lot of people ask me very similar questions along the lines of:

> How can I access tiled/ldtk data on load.

I don't mind answering those questions, but I don't think bevy_ecs_tilemap should support things like:
- Collisions
- Tiled map editor objects
- Backgrounds
- Non tile map stuff.

Which makes me think that the ldtk/tiled features should be removed in favor of using clearly defined examples. 

### Why examples?
1. Less complexity. Tiled/LDTK are level editors and as such they provide a lot more data specific to the game you are making. As such its almost impossible to support all the different ways people will use loaded data.
2. The code for the tiled/ltdk loaders are very small. Moving them into well documented example files allows users to learn how that data is loaded in and how to create the tile maps using `bevy_ecs_tilemap`.
3. If I support collisions for example it locks `bevy_ecs_tilemap` into a collision library. I would like to avoid that at all costs.

I want to reach out to the community first before I go ahead and make this change. Anyone who wants to voice any concerns/comments feel free! 

Reference issues:
#82 
#80
#21

alice-i-cecile:
Fixes #84.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,19 +11,17 @@ exclude = ["assets/*", "screenshots/*"]
 
 [features]
 default = []
-ldtk = ["anyhow", "ldtk_rust", "serde_json"]
-tiled_map = ["anyhow", "tiled"]
 
 [dependencies]
-anyhow = { version = "1.0", optional = true }
 bevy = { version = "0.5", default-features = false }
-ldtk_rust = { version = "0.5", optional = true }
 morton-encoding = "2.0"
 log = "0.4"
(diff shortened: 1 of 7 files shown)


alice-i-cecile:
The examples aren't particularly pretty or clear, but that should be fixed later IMO.

There's a lot of duplicated code in the examples in general, and they could use a cohesive pass.

Issue #103, opened 2021-10-11
Implement bevy's new render.


StarArawn:
Things that have to happen(in order):
- [x] 1. Update bevy to the pipelined-rendering branch
- [x] 2. Rewrite pipeline code for new renderer. Look into using specialized stuff that bevy added for customization.
- [x] 3. Rewrite shaders in WGSL. Preferably using some simple include functionality with string merging. This will have to be custom until bevy supports it out of the box. I think it would be straight forward enough to load in multiple shader files and then do string replacements on some key values like: `{uv_calculations}` or something similar.

StarArawn:
Work is done here: #113

PR #113, opened 2021-11-07
New renderer

StarArawn:
fixes: #103


StarArawn:
I've decide to merge this into main. If anyone would like to use a 0.5 version I would suggest using commit: `d7c4f8c8aa7775061f38aaf22bf2040610012f31`.

PR #115, opened 2021-11-12
Optimize Tiled example asset dependencies

rparrett:
I have a Tiled map with a single 16x16 tileset.

Prior to this PR, the Tiled example would create an identical `AssetPath` and add it as an asset dependency 256 times. I haven't looked into the details, but bevy was not handling this well as one would hope, and there seemed to be a noticeable slowdown.

This seems to resolve the issue.

### DIFF
--- a/examples/helpers/mod.rs
+++ b/examples/helpers/mod.rs
@@ -1,3 +1,3 @@
 pub mod camera;
-pub mod texture;
 pub mod movement;
+pub mod texture;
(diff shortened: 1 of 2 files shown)


StarArawn (approved):
Looks good! I'm holding off on merging this until the new renderer is merged. I'm thinking that I'll merge that shortly..

StarArawn:
Can you rebase/merge this so I can get it merged in?

rparrett:
Thanks for the poke. Rebased.

PR #120, opened 2021-11-30
Derive Copy on TileParent

MrGVSV:
Maybe there's a reason to not do this, but it seems like `TileParent` could easily derive `Copy` just as `Tile` and `GPUAnimated` do.


StarArawn (approved):

StarArawn:
Can you rebase/merge this so I can get it merged in?

PR #122, opened 2021-12-13
Default renderer

rparrett:
This is mostly boring mechanical work of switching includes around. I'm sure you're also working on this, but I figured I'd press the PR button just in case. Feel free to close or whatever.

The biggest change other than that that I can remember is that `Visible` seems to have been renamed to `Visibility`?

All the examples build and run and do stuff, but I didn't check that the output is identical to pre-this-pr.


StarArawn:
@rparrett Sorry this took me awhile to merge it looks good and thanks for the help!

PR #126, opened 2021-12-14
Fix panic

### LINKED ISSUE #125: new-renderer: panic with larger tilesheet
I attempted swapping in my 512x512 tilesheet into the `map` example, changing only the tile and texture size.

Curiously, using `ChunkSize(62, 62)` (or greater) fixes the panic.

I am seeing the same panic in my game, but can't seem to find a set of values that works, despite using the same tilesheet. (But `Tiled` is involved there.)

```
2021-12-13T23:43:39.803861Z  INFO bevy_render2::renderer: AdapterInfo { name: "Apple M1 Max", vendor: 0, device: 0, device_type: DiscreteGpu, backend: Metal }
2021-12-13T23:43:39.857159Z  INFO bevy_ecs_tilemap::layer: Updating tile cache.    
thread 'TaskPool (5)' panicked at 'called `Option::unwrap()` on a `None` value', /Users/robparrett/.cargo/git/checkouts/bevy_ecs_tilemap-8c720d5970222b21/2a1d51e/src/render/pipeline.rs:399:64
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'task has failed', /Users/robparrett/.cargo/registry/src/github.com-1ecc6299db9ec823/async-task-4.0.3/src/task.rs:368:45
```

repro here: https://github.com/rparrett/bevy_ecs_tilemap_test

rparrett:
Fixes #125

This seems to make some sense to me, and I came to this change by looking at the extract function in `bevy_sprite`. But I'm not really up to speed on the new renderer yet so I am not very confident that this is correct.

Either way, it fixes the panic I was seeing both in the minimal repro and in my game.

### DIFF
--- a/src/render/pipeline.rs
+++ b/src/render/pipeline.rs
@@ -67,21 +67,24 @@ pub fn extract_tilemaps(
         &TilemapUniformData,
         &Handle<Mesh>,
     )>,
+    images: Res<Assets<Image>>,
 ) {
     let mut extracted_tilemaps = Vec::new();
     for (entity, transform, chunk, tilemap_uniform, mesh_handle) in query.iter() {
-        let transform = transform.compute_matrix();
-        extracted_tilemaps.push((
-            entity,
-            (
-                LayerId(chunk.settings.layer_id),
-                chunk.material.clone(),
-                chunk.settings.mesh_type.clone(),
-                mesh_handle.clone_weak(),
-                tilemap_uniform.clone(),
-                MeshUniform { transform },
-            ),
-        ));
+        if images.get(&chunk.material).is_some() {
+            let transform = transform.compute_matrix();
+            extracted_tilemaps.push((
+                entity,
+                (
+                    LayerId(chunk.settings.layer_id),
+                    chunk.material.clone(),
+                    chunk.settings.mesh_type.clone(),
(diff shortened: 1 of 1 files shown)


src/render/pipeline.rs
```diff
@@ -71,21 +71,24 @@ pub fn extract_tilemaps(
...
-                LayerId(chunk.settings.layer_id),
-                chunk.material.clone(),
-                chunk.settings.mesh_type.clone(),
-                mesh_handle.clone_weak(),
-                tilemap_uniform.clone(),
-                MeshUniform { transform },
-            ),
-        ));
+        if images.get(&chunk.material).is_some() {
+            let transform = transform.compute_matrix();
```
turboMaCk, line 79: if I'm correct than this replaces panic with silent failure when image is not available, is that correct? I think it would be helpful to at least log some sort of warning in these cases.

StarArawn: It's quite common in bevy to silently fail in this case. If the asset fails to load the asset system will warn.

StarArawn:
@rparrett This looks good! Can you merge or rebase and I'll merge it in! :)

rparrett:
Rebased

Issue #138, opened 2022-01-02
Tile flipping is broken


StarArawn:
Should be straight forward to fix in the shader. We should also clean up the shader and remove older code.

Issue #136, opened 2022-01-01
layer_id and map_id API improvements.


turboMaCk:
One thing that confused a hell out of me were seemingly magic numbers in public API. Only after reading the source I understood how these are handled.  I think it would be a good idea to clean this API to make it easier to understand and use correctly.

From my understanding `layer_id` sort of has 2 functions at the moment:

- identifier of layer
- translation.z (z_index)

There also seems to be bug related to how layer_id is currently handled see: https://github.com/StarArawn/bevy_ecs_tilemap/issues/134. For this issue I will be ignoring the translation part.

example code snippet that could be found in many examples and likely in any sort of use-case:

```rust
// initialize map
let map_entity = commands.spawn().id();
let map_id = 0u16;
let mut map = Map::new(map_id, map_entity);

// configure layer
let layer_settings = LayerSettings::new(
    MapSize(2, 2),
    ChunkSize(8, 8),
    TileSize(16.0, 16.0),
    TextureSize(96.0, 16.0),
);

// use one of the ways to set id to the layer.... pick your preferred way...

// set layer_id using public accessor on struct
layer_settings.layer_id = 1u16;
// set layer_id using setter api of settings struct
layer_settings.set_layer_id(1u16);

// use one of the ways to set map id to the layer.... pick your preferred way...

// set layer_id using public accessor on struct
layer_settings.map_id = map_id;
// set layer_id using setter api of settings struct
layer_settings.set_map_id(map_id);

// initialize layer builder repeating layer_id and map_id (repeating what we did above)
// you can clearly see that we pass these values twice
//    - as part of layer_settings
//    - as own argumnet
let (mut layer, layer_entity) =
        LayerBuilder::new(&mut commands, layer_settings, layer_settings.map_id, layer_settings.layer_id);
        
//  use builder in some way ...

// add layer to map...  repeating map_id one more time.
map.add_layer(&mut commands, layer_settings.map_id, layer_entity);
```

[...]

PR #139, opened 2022-01-02
Cleanup and Fixes

StarArawn:
Fixes: #138 
Fixes: #136

### DIFF
--- a/src/map_query.rs
+++ b/src/map_query.rs
@@ -422,6 +422,7 @@ impl<'w, 's> MapQuery<'w, 's> {
                         grid_size.x,
                         grid_size.y,
                     );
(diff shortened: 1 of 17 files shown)


turboMaCk:
Can speak only about changes in examples but what I see makes sense! thanks.

src/map_query.rs
```diff
@@ -422,6 +422,7 @@ impl<'w, 's> MapQuery<'w, 's> {
                         grid_size.x,
                         grid_size.y,
                     );
+                    dbg!(grid_size, layer_size_in_tiles, map_size, map_pos, center);
```
subnomo, line 425: I've created a PR to revert this change: #152.

PR #162, opened 2022-03-03
Convert tile colors to linear RGBA

NeoRaider:
Tile colors must be converted to linear RGBA, otherwise dark colors will
be significantly too bright. This matches the handling of Sprite::color in
bevy_sprite.

While we're at it, also simplify the code a bit.

### DIFF
--- a/src/mesher.rs
+++ b/src/mesher.rs
@@ -70,32 +70,8 @@ impl ChunkMesher {
                         [tile_pos.x, tile_pos.y, animation_speed],
                     ]));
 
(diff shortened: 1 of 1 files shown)


NeoRaider:
I've also made a few screenshots to show the difference. For illustration purposes, this example uses a plain white texture, and the Tile::color for the "walls" is set to `Color::rgba(0.5, 0.5, 0.5, 1.0)`.

Before:
before

After:
after

StarArawn (approved):

PR #179, opened 2022-04-20
Code Compilation and Formatting Check Workflows

divark:
This includes
- Checking for compilation errors on Windows, MacOS, and Ubuntu for the whole project via cargo check. Example of results found [here](https://github.com/divark/bevy_ecs_tilemap/actions/runs/2815004984).
- Ensuring that all code adheres to a standard format via cargo fmt --check. Example of results found [here](https://github.com/divark/bevy_ecs_tilemap/runs/7717890335?check_suite_focus=true).


divark:
Currently, the CI running cargo test is having some issues with the examples provided in the documentation. Should these examples be fixed, or ignored?

StarArawn:
> Currently, the CI running cargo test is having some issues with the examples provided in the documentation. Should these examples be fixed, or ignored?

I'm not sure how I feel about clippy since it makes some code quality suggestions I'm not sure I agree with. Also it would take a bit of time to configure it how I would like. I don't think we should be running any tests since this library doesn't really have any tests worth running.

In one of my other projects I do run a light cargo format check. I could see that being helpful here:
https://github.com/StarArawn/kayak_ui/blob/main/.github/workflows/rust.yml

divark:
Okay, that sounds reasonable. Based on your feedback, I replaced the call to run tests to just check the code for all targets, and made the code quality workflow run the light cargo format check from your project. I also made the dependencies and workflow triggers consistent from the project to this CI.

How does this look for [code quality](https://github.com/divark/bevy_ecs_tilemap/actions/runs/2227930575), and for the [CI](https://github.com/divark/bevy_ecs_tilemap/actions/runs/2227930571)?

divark:
Hey @StarArawn, congratulations for getting through the recent rewrite! I was wondering if you'd still be interested in this feature.

alice-i-cecile:
We are still interested in this :) I'm going to push up a couple changes here with @bzm3r.

alice-i-cecile:
I couldn't get pushing to your branch working, so I continued this in #248. Thanks!

Issue #170, opened 2022-04-11
A Rewrite


StarArawn:
Hello author here!

I wrote an article about the state of tilemaps in bevy. Mostly around my experience with creating `bevy_ecs_tilemap` and its pitfalls. You can find it here:
https://startoaster.netlify.app/blog/tilemaps/

My main takeaway is that `bevy_ecs_tilemap` needs a re-write in order to facilitate a better API design. Feel free to leave any questions you have here!

Be on the lookout for a branch called `rewrite`. I'll comment here as well when I push it up. This branch will track the changes I make and will be a from the ground up re-write of the entire library. Feature wise I'll be focusing on the basics and moving on to add more features as I go.

What does this mean for the code base currently? I will no longer be accepting any PR's that add new features. This is to reduce the amount of work I'll end up needing to do during this re-write period. I will still accept bug fixes. 

If you would like to see any new features, or suggest changes to the new API design, feel free to leave them here!

Thanks!

StarArawn:
> Are you planning on updating the plugin in its current state for bevy 0.7?

I'm not sure when bevy 0.7 will come out. Regardless I think an update will happen.

rparrett:
We are being told perhaps tomorrow.

sizeak:
This sounds great, thanks for all the hard work you're putting into this. The most important feature to me is performant large multi-layered isometric tile maps. The only thing that comes to mind from an API perspective is keeping anything chunk related out of the public API as much as possible, it feels like an implementation leak, but I think you;re already trying to do this.

I wonder if geometry shaders still suck 7 years later?

StarArawn:
> The most important feature to me is performant large multi-layered isometric tile maps.

I have a newer isometric tile map renderer I'm using for a game I'm making. It can handle upwards of 1m tiles at 30fps. There are some downsides I've discovered:

- As bevy entity counts increase rendering performance decreases. At 40m tiles for example I was rendering at 5fps most of which was being consumed on the CPU side by bevy's internal logic. I filed a bug report here:
https://github.com/bevyengine/bevy/issues/3953 I think there's been a little bit of progress on it.
- Mapping screen coordinates to 3D tile coordinates is a nightmare. I have a somewhat working solution. 
[...]

PR #140, opened 2022-01-02
remove layer_id and map_id form `LayerSettings` struct

turboMaCk:
resolves https://github.com/StarArawn/bevy_ecs_tilemap/issues/136

## Previously:

```rust
// initialize map
let map_entity = commands.spawn().id();
let map_id = 0u16;
let mut map = Map::new(map_id, map_entity);

// configure layer
let layer_settings = LayerSettings::new(
    MapSize(2, 2),
    ChunkSize(8, 8),
    TileSize(16.0, 16.0),
    TextureSize(96.0, 16.0),
);

// set layer_id using public accessor on struct
layer_settings.layer_id = 1u16;
// set layer_id using setter api of settings struct
layer_settings.set_layer_id(1u16);

// set layer_id using public accessor on struct
layer_settings.map_id = map_id;
// set layer_id using setter api of settings struct
layer_settings.set_map_id(map_id);

// initialize layer builder
let (mut layer, layer_entity) =
        LayerBuilder::new(&mut commands, layer_settings, map_id, layer_settings.layer_id);

// add layer to map
map.add_layer(&mut commands, layer_settings.layer_id, layer_entity);
```

## After #139

```rust
// initialize map
let map_entity = commands.spawn().id();
let map_id = 0u16;
let mut map = Map::new(map_id, map_entity);

// configure layer
let layer_settings = LayerSettings::new(
    MapSize(2, 2),
    ChunkSize(8, 8),
    TileSize(16.0, 16.0),
    TextureSize(96.0, 16.0),
);

let layer_id = 1u16;

// initialize layer builder
let (mut layer, layer_entity) =
        LayerBuilder::new(&mut commands, layer_settings, map_id, layer_id);

// add layer to map
map.add_layer(&mut commands, layer_id, layer_entity);
```

## Now

```rust
// initialize map
let mut map = Map::new(&mut commands, 0u16);

// configure layer
let layer_settings = LayerSettings::new(
    MapSize(2, 2),
    ChunkSize(8, 8),
    TileSize(16.0, 16.0),
    TextureSize(96.0, 16.0),
);

// initialize layer builder
let mut layer_builder =
        map.layer_builder(&mut commands, layer_settings, 1u16);
```

~I can't run examples from main. So I relied just on a compiler errors to implement this on top of main.~

PR #180, opened 2022-04-20
Rewrite

StarArawn:
Closes: #170 
Closes: #205
Closes: #140
Closes: #121
Closes: #71


examples/random_map.rs
```diff
@@ -1,61 +1,61 @@
...
+                .insert(LastUpdate::default())
+                .insert(TileVisible(true))
+                .id();
+            tile_storage.set(&tile_pos, Some(tile_entity));
         }
-        commands
-            .entity(tile_entity.unwrap())
-            .insert(LastUpdate::default());
-    });
+    }
```
modulozero, line 39: 🙏🏻

To the completely ignorant of game dev person that I am, this is so much more intuitive. Can't wait for the post-rewrite version.

src/tiles/mod.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+#[derive(Bundle, Default, Clone, Copy, Debug)]
+pub struct TileBundle {
+    pub position: TilePos2d,
+    pub texture: TileTexture,
+    pub tilemap_id: TilemapId,
+    pub visible: TileVisible,
+    pub flip: TileFlip,
+}
```
Trouv, line 69: Is there any play to support colors on tiles, like the old `Tile` component?

src/tiles/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use bevy::{
+    math::UVec2,
+    prelude::{Bundle, Color, Component},
+};
+pub use storage::*;
+
+use crate::map::TilemapId;
+
+#[derive(Component, Default, Clone, Copy, Debug)]
```
Trouv, line 11:
```suggestion
#[derive(Component, Default, Clone, Copy, Debug, Hash)]
```
My plugin makes liberal use of `HashMap<TilePos>`

Trouv (changes requested):

src/tiles/mod.rs
Trouv: Thanks for implementing this!

rparrett:
FWIW, I migrated a project to the rewrite branch and didn't have any issues. But I'm pretty much just loading a single static Tiled map.

StarArawn:
> FWIW, I migrated a project to the rewrite branch and didn't have any issues. But I'm pretty much just loading a single static Tiled map.

That's great news! Thanks for letting me know! 🙂

src/tiles/mod.rs
StarArawn: This is done.

NeoRaider:
@StarArawn I think commit https://github.com/StarArawn/bevy_ecs_tilemap/pull/205/commits/926ceea43a617c379097819b3e0953c565be47d4 would still be a good improvement for the rewrite. Do you want to cherry-pick it, or should I open a new PR?

StarArawn:
So, according to one of Bevy's rendering devs that import is going to be removed. It was suggested to just keep what I have for right now.

NeoRaider:
I see! In that case the struct definition should be updated to https://github.com/bevyengine/bevy/blob/6752c9c59b3b81168ea6607767cbb519e5127976/crates/bevy_sprite/src/mesh2d/mesh2d_view_types.wgsl to match the Rust side of ViewUniform (this shader import was actually broken in the Bevy 0.8 release and fixed in a subsequent commit)

StarArawn:
> I see! In that case the struct definition should be updated to https://github.com/bevyengine/bevy/blob/6752c9c59b3b81168ea6607767cbb519e5127976/crates/bevy_sprite/src/mesh2d/mesh2d_view_types.wgsl to match the Rust side of ViewUniform (this shader import was actually broken in the Bevy 0.8 release and fixed in a subsequent commit)

Ideally I suppose we should use bevy's import. Until we can I think copying it into our shader is probably best.

PR #197, opened 2022-06-24
Fix Tile Bleeding When Using Atlas Feature

### LINKED ISSUE #184: visual tearing between tiles
I have a somewhat working example at https://github.com/Aposhian/bevy-sandbox/tree/tilemap-refactor which you can run with
```
cargo run --example tiled --release
```
There are tearing artifacts that show up between tiles that is variable as I move the camera around. This same issue occurred when I implemented without bevy_ecs_tilemap, and simply rendered each tile as a sprite.
image

zicklag:
Fixes #184 ! 🎉 ( I think )

This issue has plagued me in all the tile renderer's I've written over the last couple years, but I think this might finally fix it. :)

The strategy is to offset the UV sampling of the tilemap one half-pixel away from the edges of the tile so that it doesn't come close to sampling onto the next tile near the edges.

It seems to work great on my hardware, but it could be different for different hardware. Please try it out and see if it works!

### DIFF
--- a/src/render/pipeline.rs
+++ b/src/render/pipeline.rs
@@ -167,7 +167,7 @@ impl FromWorld for TilemapPipeline {
         let uniform_layout = render_device.create_bind_group_layout(&BindGroupLayoutDescriptor {
             entries: &[BindGroupLayoutEntry {
                 binding: 0,
-                visibility: ShaderStages::VERTEX,
+                visibility: ShaderStages::VERTEX | ShaderStages::FRAGMENT,
                 ty: BindingType::Buffer {
                     ty: BufferBindingType::Uniform,
                     has_dynamic_offset: true,
--- a/src/render/shaders/tilemap-atlas.wgsl
+++ b/src/render/shaders/tilemap-atlas.wgsl
@@ -28,7 +28,7 @@ var<uniform> tilemap_data: TilemapData;
 
 struct VertexOutput {
     [[builtin(position)]] position: vec4<f32>;
-    [[location(0)]] uv: vec2<f32>;
+    [[location(0)]] uv: vec4<f32>;
     [[location(1)]] color: vec4<f32>;
 };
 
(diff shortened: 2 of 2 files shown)


zicklag:
Oh, wait, I messed something up. I've got to fix it.

zicklag:
Does anybody know if branching ( using an `if` statement ) is something to really try to avoid in fragment shaders? I've seen people mention it in random places before and I kind of got the feeling that it was something you want to avoid if possible.

I have a version that works without branching, but it's harder to read and I only got it to work if I handled clamping overflow in the positive x or y direction, and not in in the negative x or y direction.

Apparently for my hardware it only bleeds in the positive direction, but I'm not sure if it's like that for all hardware.

Anyway, if the `if` statements are a problem I can try more to fix them, but if they aren't then it's easier to read as is.

StarArawn (approved):

StarArawn:
> Does anybody know if branching ( using an `if` statement ) is something to really try to avoid in fragment shaders? I've seen people mention it in random places before and I kind of got the feeling that it was something you want to avoid if possible.
> 
> I have a version that works without branching, but it's harder to read and I only got it to work if I handled clamping overflow in the positive x or y direction, and not in in the negative x or y direction.
> 
> Apparently for my hardware it only bleeds in the positive direction, but I'm not sure if it's like that for all hardware.
> 
> Anyway, if the `if` statements are a problem I can try more to fix them, but if they aren't then it's easier to read as is.
> 
> And this branch should be ready and working again.

Branching should be avoided in some cases, but I'm not too worried about it here I think. 👍

zicklag:
Awesome! Would we be able to make a new release with the fix?

I'm wanting to release my crate which depends on it, but since I use the atlas feature I wanted to get this fixed before I did the release.

StarArawn:
> Awesome! Would we be able to make a new release with the fix?
> 
> I'm wanting to release my crate which depends on it, but since I use the atlas feature I wanted to get this fixed before I did the release.

I'm not planning on releasing until the rewrite is finished. I just moved this change over to the rewrite.

zicklag:
OK, no problem. :+1:

PR #203, opened 2022-07-29
Wiring up animation!

bzm3r:
Everything was set up for animation, but what was left was connecting the dots (making sure the animation data is extracted in the extract phase and then properly packed in a `PackedTile`), and creating a relevant example in order to test it. 

This PR connects those dots.

Sadly, animation does not yet work, so I am clearly missing something. Will look into it further, but putting it out here in order to get some feedback.

### DIFF
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -22,8 +22,8 @@ fn create_background(mut commands: Commands, asset_server: Res<AssetServer>) {
 
     // To create a map we use the Tile2dStorage component.
     // This component is a grid of tile entities and is used to help keep track of individual
-    // tiles in the world. If you have multiple layers of tiles you would have a Tilemap2dStorage
-    // component per layer.
+    // tiles in the world. If you have multiple layers of tiles you would have an entity per layer,
+    // each with their own Tilemap2dStorage component.
     let mut tile_storage = Tile2dStorage::empty(tilemap_size);
 
     // Create a tilemap entity a little early.
--- a/examples/animation.rs
+++ b/examples/animation.rs
@@ -12,7 +12,7 @@ use rand::{thread_rng, Rng};
 
 mod helpers;
 
-fn create_background_tilemap(commands: &mut Commands, asset_server: &Res<AssetServer>) {
+fn create_background(mut commands: Commands, asset_server: Res<AssetServer>) {
     let texture_handle: Handle<Image> = asset_server.load("tiles.png");
 
     let tilemap_size = Tilemap2dSize { x: 20, y: 20 };
@@ -26,7 +26,7 @@ fn create_background_tilemap(commands: &mut Commands, asset_server: &Res<AssetSe
     // component per layer.
     let mut tile_storage = Tile2dStorage::empty(tilemap_size);
 
-    // Create a tilemap entity a little early
+    // Create a tilemap entity a little early.
     // We want this entity early because we need to tell each tile which tilemap entity
     // it is associated with. This is done with the TilemapId component on each tile.
     let tilemap_entity = commands.spawn().id();
@@ -66,7 +66,7 @@ fn create_background_tilemap(commands: &mut Commands, asset_server: &Res<AssetSe
         });
 }
 
-fn create_flower_tilemap(commands: &mut Commands, asset_server: &Res<AssetServer>) {
+fn create_flowers(mut commands: Commands, asset_server: Res<AssetServer>) {
     let texture_handle: Handle<Image> = asset_server.load("flower_sheet.png");
 
     let tilemap_size = Tilemap2dSize { x: 10, y: 10 };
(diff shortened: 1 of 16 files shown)


examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+                0.0,
+            ),
+            ..Default::default()
+        });
+}
+
+fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
+    commands.spawn_bundle(OrthographicCameraBundle::new_2d());
+
+    create_background_tilemap(&mut commands, &asset_server);
```
alice-i-cecile, line 134: These would be clearer and more idiomatic as their own systems IMO.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    let tilemap_size = Tilemap2dSize { x: 20, y: 20 };
+    let texture_size = Tilemap2dTextureSize { x: 96.0, y: 16.0 };
+    let tile_size = Tilemap2dTileSize { x: 16.0, y: 16.0 };
+    let grid_size = Tilemap2dGridSize { x: 16.0, y: 16.0 };
+
+    // To create a map we use the Tile2dStorage component.
+    // This component is a grid of tile entities and is used to help keep track of individual
+    // tiles in the world. If you have multiple layers of tiles you would have a Tilemap2dStorage
+    // component per layer.
```
alice-i-cecile, line 26:
```suggestion
    // tiles in the world. If you have multiple layers of tiles you would have an entity per layer,
    // each with their own Tilemap2dStorage component.
```

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+    let mut tile_storage = Tile2dStorage::empty(tilemap_size);
```
alice-i-cecile, line 27: IMO this should be called tilemap_storage.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // Create a tilemap entity a little early
+    // We want this entity early because we need to tell each tile which tilemap entity
+    // it is associated with. This is done with the TilemapId component on each tile.
+    let tilemap_entity = commands.spawn().id();
```
alice-i-cecile, line 32: The connection between this line and the component above is not obvious when first reading.

I'd swap the order, and then explicitly connect them.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+fn create_flower_tilemap(commands: &mut Commands, asset_server: &Res<AssetServer>) {
+    let texture_handle: Handle<Image> = asset_server.load("flower_sheet.png");
+
+    // Size of the tile map in tiles.
+    let tilemap_size = Tilemap2dSize { x: 10, y: 10 };
+    let texture_size = Tilemap2dTextureSize { x: 32.0, y: 448.0 };
+    let tile_size = Tilemap2dTileSize { x: 32.0, y: 32.0 };
+    let grid_size = Tilemap2dGridSize { x: 16.0, y: 16.0 };
+
+    // To create a map we use the Tile2dStorage component.
```
alice-i-cecile, line 78: These comments shouldn't be duplicated in the same example.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+use bevy_ecs_tilemap::tiles::{AnimatedTile, Tile2dStorage, TileBundle, TilePos2d, TileTexture};
+use bevy_ecs_tilemap::{Tilemap2dPlugin, TilemapBundle};
+use rand::{thread_rng, Rng};
+
+mod helpers;
+
+fn create_background_tilemap(commands: &mut Commands, asset_server: &Res<AssetServer>) {
+    let texture_handle: Handle<Image> = asset_server.load("tiles.png");
+
+    let tilemap_size = Tilemap2dSize { x: 20, y: 20 };
```
alice-i-cecile, line 18: These should be consts.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+
+    // Create a tilemap entity a little early
+    // We want this entity early because we need to tell each tile which tilemap entity
+    // it is associated with. This is done with the TilemapId component on each tile.
+    let tilemap_entity = commands.spawn().id();
+
+    let mut random = thread_rng();
+    // Spawn the elements of the tilemap.
+    for _ in 0..10 {
+        let tile_pos = TilePos2d {
```
alice-i-cecile, line 92: This is a poor randomization strategy, as it allows for overlaps. You want to be sampling without replacement.

examples/animation.rs
```diff
@@ -0,0 +1,154 @@
...
+            .spawn()
+            .insert_bundle(TileBundle {
+                position: tile_pos,
+                tilemap_id: TilemapId(tilemap_entity),
+                texture: TileTexture(0),
+                ..Default::default()
+            })
+            .id();
+        tile_storage.set(&tile_pos, Some(tile_entity));
+        commands.entity(tile_entity).insert(AnimatedTile {
```
alice-i-cecile, line 106: This needs to be called out clearly with comments; it's the whole point of the example and it's easy to miss.

StarArawn (approved):

PR #211, opened 2022-08-07
Impl PartialEq and PartialOrd on TilePos

tehsmeely:
Adding PartialEq and PartialOrd as derives on TilePos because it's pretty common to want to do a bunch of operations on TilePos

PartialEq certainly at least, could take of leave PartialOrd as the default lexicographic ordering might not make sense
(i.e. that TilePos{x:2, y:0} > TilePos {x:1, y:100})


StarArawn (approved):
Makes sense and is a simple change! 🙂

tehsmeely:
Thanks!

PR #215, opened 2022-08-10
Fix grid size: scale tile position by grid_size first, then, then add tile_size to get the corners

bzm3r:
Currently, the `fix-grid-size` branch scales tile positions by `tile_size`, but this does not place the tiles at `grid_size` multiples (rather it places them at `tile_size` multiples). So, first we should scale by `grid_size`, and then add rectangular `tile_size`d offsets to get the corner positions.


StarArawn:
closes: #214

StarArawn (approved):

PR #219, opened 2022-08-11
General fixes to stability.

StarArawn:


mbolt35 (commented):
This seems to have resolved the reduced frame rate! There does seem to be a small leak that occurs, but's not nearly as significant as before. 

Any of the procedurally generated tilemaps I had just required adding the:
```
commands.entity(tile_map_entity).add_child(tile_entity)
```

during generation. 

I haven't explored the performance implications much, but is there a way to batch add children? 

From a usability standpoint, this definitely seems easy to get tripped up on, but it's somewhat unclear how to best include the add_child as part of the tilemap API. 

Thank you for tracking this down, I'll keep my eye out on those bevy PRs as well.

StarArawn:
> I haven't explored the performance implications much, but is there a way to batch add children?
Not yet, but I think we can add this. 👍

Issue #58, opened 2021-06-28
Layout of hex tiles is inaccurate, with error accumulating


alice-i-cecile:
While working on a tile-coordinates-from-mouse-click utility, @plof27 and I noticed that coordinates of the hex tiles become progressively more inaccurate the farther from the origin you get. You can see the [utility function](https://github.com/Leafwing-Studios/bevy-tower-defense-template/blob/main/src/utils.rs#L39) we used for conversions at the link.

We noticed this using a point-up, `Hextype::Row` layout. It works perfectly for x, but gets progressively worse on the y-axis as you get further from the origin. I suspect this is a layout bug as a result; particularly as it was still an issue using an alternate pixel-based renderer.

StarArawn:
What map size does it start to become inaccurate at? Do you have a screenshot I can see of the inaccuracies?

alice-i-cecile:
I'll see if I can grab some more diagnostics for you soon. If you need it sooner, the full repo linked there is open source and you can poke around.

bzm3r:
Fixed by #268

PR #228, opened 2022-08-16
Reworking `TilemapMeshType`, fixing iso staggered rendering, and adding helper functions to return neighboring directions of a tile.

bzm3r:
`TilemapMeshType` has been reworked to `TilemapType`. 

Helper functions have been added/updated to return neighbors, and to get the position of a tile in world space. 

Bugs in rendering of staggered iso have been fixed.

In the process of fixing the staggered iso bugs, I have also "standardized" how iso diamond works. Maybe this is a change that is not wanted, in which case I can revert it.


src/tiles/storage.rs
```diff
@@ -1,26 +1,37 @@
...
+use crate::TilemapMeshType;
 
 use super::TilePos;
 
 /// Used to store tile entities for fast look up.
 /// Tile entities are stored in a grid. The grid is always filled with None.
 #[derive(Component, Default, Debug, Clone)]
-pub struct TileStorage {
+pub struct TilemapStorage {
+    mesh_type: TilemapMeshType,
```
StarArawn, line 12: Storing `TilemapMeshType` here means that we hard lock on requiring a `TilemapStorage`. Although `TilemapStorage` being optional is not very well documented.

src/render/extract.rs
```diff
@@ -120,14 +121,11 @@ pub fn extract(
         Query<
             Entity,
             Or<(
-                Added<TilemapMeshType>,
```
StarArawn, line 123: A lack of an `Added` query filter here seems strange to me as we wont detect when a tile map initially spawns?

examples/hexagon_row.rs
```diff
@@ -59,23 +60,22 @@ fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
...
             ..Default::default()
         });
 }
 
-fn swap_mesh_type(mut query: Query<&mut TilemapMeshType>, keyboard_input: Res<Input<KeyCode>>) {
+fn swap_mesh_type(mut query: Query<&mut TilemapStorage>, keyboard_input: Res<Input<KeyCode>>) {
     if keyboard_input.just_pressed(KeyCode::Space) {
-        for mut tilemap_mesh_type in query.iter_mut() {
-            match *tilemap_mesh_type {
+        for mut tilemap_storage in query.iter_mut() {
```
StarArawn, line 69: Does this even do anything anymore? From what I can tell the mesh type can't be changed.

bzm3r: Yes, it does do something, because the mesh type can be changed because we provide a `set_mesh_type` method to change the mesh type. Then, when `extract` queries for `TilemapStorage`, it calls `get_mesh_type` on `TilemapStorage` which returns a different value.

src/render/extract.rs
bzm3r: @StarArawn I guess we should have an `Added<TilemapStorage>` here...

StarArawn (approved):

StarArawn (changes requested):
Some of the examples don't work I think they just need to be updated with the new types?

StarArawn:
Fixes: #231

StarArawn:
@bzm3r Can you make sure I didn't miss any issues this fixes? Or possibly ones it doesn't address completely. Thanks!

Affinator:
5 layers with TilemapSize { x: 1, y: 65 };
grafik

with TilemapSize { x: 65, y: 16 };
grafik

So it seems to be a problem with a TileMapHeight above 64.

Affinator:
I created a pull request with a fix (possibly) https://github.com/Leafwing-Studios/bevy_ecs_tilemap/pull/1 

Please check if my fix is doing it correctly or is just working by accident.

bzm3r:
@Affinator can you confirm this issue is fixed for you using the latest version of this PR?

bzm3r:
@StarArawn I think the issue list you have is correct, except for #58 --- I have not touched anything to do with converting pixel coordinates to hex cells, and I also did not touch hex layout. (I mainly touched code that converts a tile position into a pixel coordinates, and how neighbours are reported for hexes.) I think investigating #58 will require a separate analysis on my part.

src/render/extract.rs
```diff
@@ -120,14 +120,14 @@ pub fn extract(
...
-                Changed<TilemapMeshType>,
+                Added<TilemapType>,
+                Changed<TilemapType>,
                 Changed<GlobalTransform>,
                 Changed<TilemapTexture>,
                 Changed<TilemapTileSize>,
                 Changed<TilemapSpacing>,
                 Changed<TilemapGridSize>,
-                Changed<TilemapMeshType>,
+                Changed<TilemapType>,
```
StarArawn, line 130: This is a duplicate of line 124, I'm actually surprised bevy doesn't complain! In any case can we remove one of these change filters?

StarArawn (changes requested):

StarArawn:
This is looking good! I think we just need one more change and it's ready for merge. 👍

examples/colors.rs
```diff
@@ -3,23 +3,34 @@ use bevy_ecs_tilemap::prelude::*;
...
 
-fn fill_tilemap_rect_color(
+/// This is like `fill_tilemap_rect`, a function in `helpers.rs`, except that it
+/// also sets a color.
+///
+/// Fills a rectangular region with the given tile.
+///
+/// The rectangular region is defined by an `origin` in `TilePos`, and a size
+/// in tiles (`TilemapSize`).
+pub fn fill_tilemap_rect_color(
```
StarArawn, line 13: Perhaps we should move this to helpers?

StarArawn (changes requested):
One last change. 😓 It's looking really good though!

StarArawn (approved):

Affinator:
iso_staggered is still broken in this PR as well as in main (although with different render errors).

Affinator:
Interesting: If I set the anchor for the chunks to (0,0) in helpers.rs -> get_chunk_2d_transform

    let anchor_tile_pos = TilePos {
        x: 0,    // chunk_position.x * chunk_size.x,
        y: 0,    // chunk_position.y * chunk_size.y,
    };

then iso_staggered renders correctly.. ?

PR #232, opened 2022-08-24
Update chunk position with scale

qazzz789:
* Fixes Chunks artifact when changing the scale of a map
* Tested with other map types, hex and iso

image


StarArawn (approved):
This looks correct but will likely need a rebase/merge as I will be merging #228 first.

qazzz789:
Sounds good, ill do another test after the rebase to make sure its still valid.

PR #240, opened 2022-08-31
Fixed remaining issue with staggered iso rendering.

bzm3r:
Something left over from the testing/exploratory phase in the iso_staggered shader was causing an issue.

### DIFF
--- a/examples/hexagon_row.rs
+++ b/examples/hexagon_row.rs
@@ -3,7 +3,7 @@ use bevy_ecs_tilemap::prelude::*;
 mod helpers;
 
 // Side length of a colored quadrant (in "number of tiles").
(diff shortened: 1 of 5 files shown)


examples/hexagon_row.rs
```diff
@@ -3,7 +3,7 @@ use bevy_ecs_tilemap::prelude::*;
 mod helpers;
 
 // Side length of a colored quadrant (in "number of tiles").
-const QUADRANT_SIDE_LENGTH: u32 = 32;
+const QUADRANT_SIDE_LENGTH: u32 = 80;
```
StarArawn, line 6: Not really an issue but this produces 3x3 render chunks. Probably best for testing.

StarArawn (approved):

PR #246, opened 2022-09-03
Add TilemapRenderSettings Resource to set Custom Chunk Size

mbolt35:
This change came from some testing I was doing in my game with different chunk sizes. It seemed reasonable that for larger tile sizes, it may be preferable to reduce the chunk size. It _seemed_ to yield slightly better performance compared to the default 64x64 in my case (I didn't use any rigorous benching to verify this, just basic FPS observations). 

I'm not sure if this is something that would be useful, but I figured I'd clean up my test code and offer up the PR just in case. Of course, no worries if it's an undesirable change -- feel free to ignore. :)

---- 

This change adds a resource that could contain "custom parameters" (currently just `chunk_size: UVec2`) into the render pipeline. Similar to `WindowDescriptor` has to be added before the `DefaultPlugins` are added for bevy, this resource must also be added prior to the `TilemapPlugin` being added. 

For example: 
```rust
fn main() {
    App::new()
        .insert_resource(WindowDescriptor {
            width: 1270.0,
            height: 720.0,
            title: String::from(
                "Custom Chunk Size Example - Press Space to change Texture and H to show/hide tilemap.",
            ),
            ..Default::default()
        })
        // Thie example demonstrates configuring the Tilemap rendering pipeline to use a smaller
        // chunk size per mesh. It must be added before the `TilemapPlugin`.
        .insert_resource(TilemapRenderSettings {
            chunk_size: UVec2::new(32, 32),
        })
        .insert_resource(ImageSettings::default_nearest())
        .add_plugins(DefaultPlugins)
        .add_plugin(LogDiagnosticsPlugin::default())
        .add_plugin(FrameTimeDiagnosticsPlugin::default())
        .add_plugin(TilemapPlugin)
        .add_startup_system(startup)
        .add_system(helpers::camera::movement)
        .add_system(swap_texture_or_hide)
        .run();
}
```

### DIFF
--- a/examples/chunk_size.rs
+++ b/examples/chunk_size.rs
@@ -1,0 +1,118 @@
+use bevy::{
+    prelude::*, 
+    diagnostic::{FrameTimeDiagnosticsPlugin, LogDiagnosticsPlugin},
+    render::texture::ImageSettings, window::PresentMode
+};
+
+use bevy_ecs_tilemap::prelude::*;
+mod helpers;
+
+fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
+    commands.spawn_bundle(Camera2dBundle::default());
+
+    let texture_handle: Handle<Image> = asset_server.load("tiles.png");
+
+    let tilemap_size = TilemapSize { x: 128, y: 128 };
+
+    // Create a tilemap entity a little early.
+    // We want this entity early because we need to tell each tile which tilemap entity
+    // it is associated with. This is done with the TilemapId component on each tile.
+    // Eventually, we will insert the `TilemapBundle` bundle on the entity, which
+    // will contain various necessary components, such as `TileStorage`.
+    let tilemap_entity = commands.spawn().id();
+
+    // To begin creating the map we will need a `TileStorage` component.
+    // This component is a grid of tile entities and is used to help keep track of individual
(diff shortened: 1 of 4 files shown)


src/render/mod.rs
```diff
@@ -130,6 +173,7 @@ impl Plugin for TilemapRenderingPlugin {
 
         let render_app = app.sub_app_mut(RenderApp);
         render_app
+            .insert_resource(ChunkCoordinates::new(chunk_size))
```
StarArawn, line 176: I would prefer if this said `RenderChunkCoordinates` as chunks are very much an internal thing right now and only a way for the renderer to split large tilemaps into more manageable data sizes.

examples/chunk_size.rs
```diff
@@ -0,0 +1,118 @@
+use bevy::{
```
StarArawn, line 1: Perhaps we can combine this example with the existing chunking example?

StarArawn (changes requested):

StarArawn:
As for performance here it depends on the following:
1. A tilemap who updates frequently will benefit from having smaller chunks.
2. A tilemap that does not update frequently will render faster with larger chunks.

It might be nice to include this advice on the resource. 👍

PR #252, opened 2022-09-04
Add the ability to configure the chunk size.

bzm3r:
Thanks to @mbolt35! 

With edits suggested by StarArawn. This PR used to be #246


StarArawn:
Closes: #246

examples/chunking.rs
```diff
@@ -37,7 +38,7 @@ fn spawn_chunk(commands: &mut Commands, asset_server: &AssetServer, chunk_pos: I
         .entity(tilemap_entity)
         .insert_bundle(TilemapBundle {
             grid_size: TILE_SIZE.into(),
-            size: CHUNK_SIZE,
+            size: CHUNK_SIZE.into(),
```
StarArawn, line 41: I think this might be confusing. It's best to not equate render chunks with user built tile map chunks here. 

I think what I would like to see is something like:

```rust
const CHUNK_SIZE: UVec2 = UVec2 { x: 32, y: 32 };

// Render chunk sizes are set to 4 render chunks per user specified chunk.
TilemapRenderSettings {
    chunk_size: UVec2::new(16.0, 16.0),
}
```

I'm open to suggestions here too though. 👍

StarArawn (changes requested):

mbolt35:
@bzm3r @StarArawn I'm going to close #246 - Thanks for the feedback and changes, this looks great to me!

src/map/mod.rs
```diff
@@ -3,6 +3,32 @@ use bevy::{
     prelude::{Component, Entity, Handle, Image},
 };
 
+/// Custom parameters for the render pipeline.
+///
+/// It must be added before [`TilemapPlugin`](crate::TilemapPlugin). For example:
+/// ```no_run
+/// App::new()
```
StarArawn, line 10: I think you have to add imports to the test or it wont compile correctly?

StarArawn (approved):

PR #258, opened 2022-09-04
Don't use bevy_render feature which includes a bunch of 3d stuff

rparrett:
Noticed my game preparing skinned meshes and lights and stuff and tracked it down here.

I don't think there's anything in `bevy_render` actually required here, except for `bevy_core_pipeline` which is already being included.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ atlas = []
 render = []
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


rparrett:
Not clear to me yet whether I'm super wrong or there's something odd going on with the clippy CI.

rparrett:
Yep, my bad. Sorry for the noise.

PR #264, opened 2022-09-06
correct entity comparisons when determining chunk visibility

### LINKED ISSUE #259: Latest RenderLayers Support Causes Issues
It appears that this pull request: https://github.com/StarArawn/bevy_ecs_tilemap/pull/244 has caused some adverse effects with the dynamic chunking approach. A lot of the chunks aren't being rendered after being added. I checked a lot of the recently added commits in main -- this commit was the one was seemed to break. 

Reading the PR, there is mention of `.insert(RenderLayers.layer(2))` -- is this a requirement on added tilemap entities? 

This is easily reproducible in the `chunking` example in the `main` branch. 
https://user-images.githubusercontent.com/334480/188558169-eb746c67-cfe3-443d-ab9b-c897b1f43ace.mp4

@Piefayth @StarArawn @bzm3r Any thoughts?

Piefayth:
Resolves #259

`contains` does not compare entities as expected. we can just compare the entity ids directly

### DIFF
--- a/src/render/queue.rs
+++ b/src/render/queue.rs
@@ -136,7 +136,11 @@ pub fn queue_meshes(
                 .unwrap();
 
             for (entity, chunk_id, transform, tilemap_id) in standard_tilemap_meshes.iter() {
-                if !visible_entities.entities.contains(&tilemap_id.0) {
+                if !visible_entities
+                    .entities
+                    .iter()
+                    .any(|&entity| entity.id() == tilemap_id.0.id())
+                {
                     continue;
                 }
 


Piefayth:
Having seen the chunking example, I'm also curious if anyone's use case is hit by the performance implications of the PR due to large numbers of tilemaps.

If they do, the upgrade that could be made here is to steal [bevy's approach](https://github.com/bevyengine/bevy/blob/59bf3c4cc968f0725aa6b97b0a0bf42b280c8b6f/crates/bevy_sprite/src/render/mod.rs#L428): for each view, pre-compute a `FixedBitSet` where each element represents a given tilemap's boolean visibility in the view. Then for each render chunk, do the lookup in the set rather than iterating the visible entities list again.

StarArawn (approved):

mbolt35:
> Having seen the chunking example, I'm also curious if anyone's use case is hit by the performance implications of the PR due to large numbers of tilemaps.

Thanks for the fix! Generally, in the "chunked" approach, the number of tilemaps loaded at any given time should be reduced based on (but not limited to) the visibility. 

The code in the PR seems to be bound to the `length` of the `visible_entities`, which I would assume would "degrade" performance for small chunk sizes (where the number of visible tilemaps was high). Is that correct? If so, given that the render chunk_size is now configurable and the developer has options to tune, my opinion is that your change is fine :)

> If so, the upgrade that could be made here is to steal [bevy's approach](https://github.com/bevyengine/bevy/blob/59bf3c4cc968f0725aa6b97b0a0bf42b280c8b6f/crates/bevy_sprite/src/render/mod.rs#L428): for each view, pre-compute a `FixedBitSet` where each element represents a given tilemap's boolean visibility in the view. Then for each render chunk, do the lookup in the set rather than iterating the visible entities list again.

I'm not certain, but think that the tilemap visibility would be excluded from this iteration. This is based on the assumption that the visible flag on tilemap excludes it from making it past the prepare stage:

https://github.com/StarArawn/bevy_ecs_tilemap/blob/5b7d8876e7dfc017bdbedbc08d5fd306b99bdead/src/render/prepare.rs#L153-L156

I'm working on a project now that can potentially increase the total number of tilemaps universally, so I will keep this concern in mind. Thanks again :)

Issue #40, opened 2021-06-01
Feature request: mouse-to-tile events


mattdm:
It would be nice to have mouse events translated to tiles. At the minimum, a helper function which takes mouse coordinates and returns a tile. However, even better, a system which:

* Generates custom events for:
    - a tile was clicked (left/right/middle/other, optionally subposition)
    - hover-in
    - hover-out
    - maybe a mouse-drag event which gives a list of tiles from first (tile where the button was pressed down) + all the others passed over in order + last (tile where button was released)
    - maybe also mouse-down / mouse-up on a tile (not sure how useful that really is vs "click" outside of the drag case)
* Provides a way to query:
    - last-clicked tile (or, list of tiles clicked since last query?)
    - currently-hovered tile
    - subposition within tile of last click (scaled to tile size in view)
    - subposition within tile of hover
    - maybe the latest mouse-drag list?
    
I'm not sure how -- or if at all -- this should interact with non-tile sprites or other UI elements. I can image cases where it'd be nice to at least know that a click is "obstructed" by an object which is above the map.

StarArawn:
Related to #30

mattdm:
And just a few more complications:

* Should handle hex and isometric tiles 
* Layers... not sure how that should work at all

mattdm:
Also connected to #44 -- if isometric tiles have _height_, they can extend out of their nominal space, and therefore mouse-picking should be based on render height. (Including the possibility of layers having their own Z axis.) But that adds a lot of complexity that I would love to not have block the basics.

alice-i-cecile:
Implementing this logic by hand for hexagonal tiles is a *giant* pain: getting this working for hex tiles will make the library substantially more useful.

plof27:
Definitely not a full solution, but this might be helpful to some. I wrote some helper functions that can convert a world position to a tile coordinate for axial hex tiles (`HexType::Row`). 
```rust
/// Returns the tile coordinates at the given world coordinates
///
/// Based on the pixel-to-hex algorithm described here: https://www.redblobgames.com/grids/hexagons/#pixel-to-hex
pub fn tile_from_world_coordinates(
    world_position: Vec2, 
    tile_size: Vec2
) -> Option<UVec2> {
    
    // Basis vectors for axial hex coordinates
    let hex_basis = Mat2::from_cols(Vec2::new(3f32.sqrt()/3.0, 0.0), Vec2::new(-1.0/3.0, 2.0/3.0));
[...]

Issue #30, opened 2021-05-25
Feature Request: API to translate map coords to screen coords


chrisburnor:
For action games, the sprite entities would be 'free-floating' entities rendered on top of the map, but the game would still want to track their position relative to the map itself.  It probably doesn't make sense to have the map *store* those entities or references to them, but it would be useful to have the map translate a position (given as a `DVec2`) and return a `Transform` or `DVec2` object that corresponds to the screen coordinates of that location given the map projection and scale and vice versa.

StarArawn:
Yes, this would be useful. I can provide an API like:
```rust
let map_pos: UVec2 = map_query.pixel_pos_to_map_pos(Vec2::new(0.0, 0.0));
assert!(map_pos == UVec2::new(0, 0));

// Assuming tiles are 16x16.
let pixel_pos: Vec2 = map_query.map_pos_to_pixel_pos(UVec2::new(0, 0), true); // boolean represents centered or not?
assert!(map_pos == Vec2::new(8.0, 8.0)); 
```

chrisburnor:
That looks good! Thanks for the fast response!

Is the `0u16` argument the layer_id? Makes sense to keep that.  Also, I'd recommend using an enumeration instead of a boolean for the third 'offset' argument. You could more easily extend it to different options (center, upper left, fixed offset, etc).  I'd also suggest having the reverse mapping of pixel to map (useful for mouse selection).

bzm3r:
Fixed by #268

PR #268, opened 2022-09-13
Helpers overhaul

bzm3r:
Per discussions, an overhaul of helpers that also with the following issues:

* https://github.com/StarArawn/bevy_ecs_tilemap/issues/30

* https://github.com/StarArawn/bevy_ecs_tilemap/issues/58 

* https://github.com/StarArawn/bevy_ecs_tilemap/issues/278

It provides the ground work for dealing with (but does not yet generate mouse events):
* https://github.com/StarArawn/bevy_ecs_tilemap/issues/40

What remains before the PR can move out of draft stage:

- [x] docs need to be updated/cleaned up (there are a lot of docs around that no longer belong where they are, or need to be simplified)

- [x] update hex shader code to use new tile position to world position code for hexagons

- [x] create an example that uses the new code, and confirms that world position -> tile position works correctly 

An extension to this PR would create mouse events as requested in #40


bzm3r:
Brief overview:

* helpers have been re-organized so that they are not in a single file

* for hex grids: added support for conversion between axial coordinate system (the coordinate system used by `Row` and `Column` hex coord system) and offset coordinate systems (`RowEven`, `RowOdd`, `ColumnEven`, `ColumnOdd`) 

* for hex grids: added support for conversion between axial coodinate system (the coordinate system used by `Row` and `Column` hex coord system) and cube coordinate systems 

* for hex grids: added support for conversion between axial/cube/offset coordinates and `TilePos`

* for hex grids: note that all these conversions between coordinate systems are cheap and require only a couple of arithmetic operations at most (maybe I should be adding the `#[inline]` hint everywhere, but I'll make a future issue for that)

* for hex grids: supporting cube positions allows us to convert a `Vec2` world space position into a `TilePos`; suppose you are using `RowOdd` coordinate system, and you have a `world_pos`, then: `world_pos` is mapped to a `FractionalAxialPos`, which is then rounded (requires conversion to `FractionalCubePos` at this stage) into an `AxialPos` (this rounding is actually the heavyweight bit), which is then converted into a `RowOddPos`, which is converted into a result `TilePos`

* for isometric grids: simplified how the math around tile layout by capturing the square -> isometric projection as a matrix, this allows us to apply a matrix inverse to easily calculate which tile a `Vec2` world position belongs to (essentially, the squishes the isometric grid back to a square grid) ---- this process is also quite similar to what is happening for hex grids, so there is some unification of code between hex grid/iso grid shaders!

* for squares/isometric grids: tile center is now at `(0, 0)`

* a lot of the shaders are now quite similar, because ultimately, the math behind them has deep similarities 

* for neighbor helpers: made lots of private functions public because https://github.com/StarArawn/bevy_ecs_tilemap/issues/278

examples/mouse_to_tile.rs
```diff
@@ -0,0 +1,383 @@
...
+            for mut section in tile_text.sections.iter_mut() {
+                section.style.color = Color::BLACK;
+            }
+            commands
+                .entity(highlighted_tile_entity)
+                .remove::<HighlightedLabel>();
+        }
+    }
+
+    for window in windows.iter() {
```
StarArawn, line 342: This block of code is probably the most important bit in the entire file for showing people how tile picking works. I think it would be helpful to explain what is happening here with some comments? 

For example:
```rust
// First we grab the mouse world position. This is done by projecting the cursor position into camera space.
if let Some(cursor_world_pos) = cursor_pos_as_world_pos(window, &camera_q) {
// We need to make sure that the cursors world position is correct due to any map transformation.
// This example only shows how to correctly modify the position based on translation
// but additional code would be needed for transforming if the tilemap's scale was changed.
let map_translation = map_transform.translation.truncate();
let world_pos = cursor_world_pos - map_translation;
// Once we have a world position we can transform it into a possible tile position. 
if let Some(tile_pos) = TilePos::from_world_pos(&world_pos, map_size, grid_size, map_type)
```

examples/mouse_to_tile.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+#[derive(Component)]
+struct HighlightedLabel;
+
+pub fn cursor_pos_as_world_pos(
+    window: &Window,
+    camera_q: &Query<(&Transform, &Camera)>,
+) -> Option<Vec2> {
+    window.cursor_position().map(|cursor_pos| {
```
StarArawn, line 305: I think we should likely pass cursor position directly into this using bevy events. It's a bit more code to write out but I think it's the recommended way of dealing with mouse input in bevy. See:
https://github.com/bevyengine/bevy/blob/main/examples/input/mouse_input_events.rs

src/helpers/geometry.rs
```diff
@@ -0,0 +1,22 @@
+use crate::{TilemapSize, TilemapTileSize, Transform};
+use bevy::math::Vec2;
+
+/// Calculates a [`Vec2`] position for a tilemap so that when set to this position, it shows up
+/// centered on the screen.
+pub fn get_tilemap_center(size: &TilemapSize, tile_size: &TilemapTileSize) -> Vec2 {
```
StarArawn, line 6: If our tiles are centered now do we need this? Or is that only the case for hex/iso?

StarArawn (changes requested):

src/helpers/geometry.rs
bzm3r: I am not sure, but I think so? This is for centering the whole tilemap, right? Currently, our tiles are "centered" in the sense that the center of a tile with `TilePos(0, 0)` lies at `[0, 0]` in world space.

StarArawn (approved):

Issue #241, opened 2022-09-01
move tile by change TilePos, old tile doesn't clean.


alexniver:
Want move tile by change tile_pos, but old tile doesn't clean.

code :

```rust
use bevy::{prelude::*, render::texture::ImageSettings};
use bevy_ecs_tilemap::prelude::*;
mod helpers;

#[derive(Component)]
struct TileMoveable;

fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
    let texture_handle: Handle<Image> = asset_server.load("tiles.png");

    let tilemap_size = TilemapSize { x: 32, y: 32 };

    let tilemap_entity = commands.spawn().id();

    let mut tile_storage = TileStorage::empty(tilemap_size);

    // Spawn the elements of the tilemap.
    for x in 0..32u32 {
        for y in 0..32u32 {
            let tile_pos = TilePos { x, y };
            let tile_entity = commands
                .spawn()
                .insert_bundle(TileBundle {
                    position: tile_pos,
                    tilemap_id: TilemapId(tilemap_entity),
                    ..Default::default()
                })
                .id();
            tile_storage.set(&tile_pos, Some(tile_entity));
        }
    }

    let tile_size = TilemapTileSize { x: 16.0, y: 16.0 };

    commands
        .entity(tilemap_entity)
        .insert_bundle(TilemapBundle {
            grid_size: TilemapGridSize { x: 16.0, y: 16.0 },
            size: tilemap_size,
            storage: tile_storage,
            texture: TilemapTexture(texture_handle),
            tile_size,
            transform: bevy_ecs_tilemap::helpers::get_centered_transform_2d(
                &tilemap_size,
                &tile_size,
                0.0,
            ),
            ..Default::default()
        });
    commands.spawn_bundle(Camera2dBundle::default());
}

fn startup_moveable_tiles2(mut commands: Commands, asset_server: Res<AssetServer>) {
    let texture_handle: Handle<Image> = asset_server.load("tiles2.png");

    let tilemap_size = TilemapSize { x: 32, y: 32 };

    let tilemap_entity = commands.spawn().id();

    let mut tile_storage = TileStorage::empty(tilemap_size);

    let tile_pos = TilePos { x: 16, y: 16 };
    let tile_entity = commands
        .spawn()
        .insert_bundle(TileBundle {
            position: tile_pos,
            tilemap_id: TilemapId(tilemap_entity),
            ..Default::default()
        })
[...]

PR #271, opened 2022-09-15
Allow users to update tile positions without despawning.

StarArawn:
Fixes: #241

### DIFF
--- a/examples/move_tile.rs
+++ b/examples/move_tile.rs
@@ -60,7 +60,7 @@ fn main() {
             width: 1270.0,
             height: 720.0,
             title: String::from(
-                "Basic Example - Press Space to change Texture and H to show/hide tilemap.",
+                "Update tile positions without despawning.",
             ),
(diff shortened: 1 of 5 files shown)


examples/move_tile.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+fn main() {
+    App::new()
+        .insert_resource(WindowDescriptor {
+            width: 1270.0,
+            height: 720.0,
+            title: String::from(
+                "Basic Example - Press Space to change Texture and H to show/hide tilemap.",
```
bzm3r, line 63:
```suggestion
                "Update tile positions without despawning.",
```

bzm3r (changes requested):
Looks good to me! Maybe we should rename the example title, but that's about it.

bzm3r (commented):
Looks good to me!

bzm3r (approved):

Issue #269, opened 2022-09-14
Make sure `uv`s for the texture array-based fragment shader are not interpolated 


bzm3r:
Currently, we rely on casting `uv.z` as `i32`. Does this have any weird edge cases? Better thing would be to add in a flag on the `uv` that specifies it should not be interpolated. See here for an example of how to do this: https://github.com/bzm3r/doug_renderers/blob/428a400d85ba989e5f78d5d7b4eb9686334bc12c/src/shaders/vpull.wgsl#L43

StarArawn:
Turns out this is indeed an issue. It happens when zooming the projection matrix in and out sometimes a tile uv.z can end up being 0.9999 because of interpolation and since we cast to i32 it drops the decimals leaving us with zero instead of one.

PR #279, opened 2022-09-20
Fixes issue where interpolation can cause the wrong tile to be rendered.

StarArawn:
Fixes: #269

### DIFF
--- a/src/render/shaders/tilemap.wgsl
+++ b/src/render/shaders/tilemap.wgsl
@@ -34,8 +34,9 @@ var<uniform> tilemap_data: TilemapData;
 #include 0
 
 struct VertexOutput {
-    @location(0) uv: vec3<f32>,
+    @location(0) uv: vec2<f32>,
     @location(1) color: vec4<f32>,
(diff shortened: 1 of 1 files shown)


src/render/shaders/tilemap.wgsl
```diff
@@ -34,8 +34,9 @@ var<uniform> tilemap_data: TilemapData;
 #include 0
 
 struct VertexOutput {
-    @location(0) uv: vec3<f32>,
+    @location(0) uv: vec2<f32>,
     @location(1) color: vec4<f32>,
+    @location(2) @interpolate(flat) tile_id: f32,
```
StarArawn, line 39: Optionally we could just use a u32 here. 🤷

inodentry: yeah why not use integers? that seems like the best way to ensure everything is correct

bzm3r (approved):
These changes look good to me! I especially like that we're not using generic `x, y, z, w` components to refer to stuff like `tile_id` anymore. Makes the code much easier to read.

Issue #277, opened 2022-09-18
Texture filtering mode is forced to Nearest, no way to use Linear


inodentry:
After a quick investigation, it seems like you are hardcoding `FilterMode::Nearest` on line 282 in `src/render/mod.rs`.

Regardless of how I have configured my original Image asset, you create your texture array with your own sampler with Nearest filtering.

I want Linear filtering for my tilemap, and currently there is no way for me to override that.

StarArawn:
We should use bevy's settings here instead. 👍

bzm3r:
Closed by #280

PR #280, opened 2022-09-20
Added a texture atlas to array texture preloader.

StarArawn:


src/array_texture_preload.rs
```diff
@@ -0,0 +1,70 @@
...
+impl ArrayTextureLoader {
+    pub fn add(&self, texture: TilemapArrayTexture) {
+        if let Ok(mut textures) = self.textures.try_write() {
+            textures.push(texture);
+        }
+    }
+
+    pub(crate) fn drain(&self) -> Vec<TilemapArrayTexture> {
+        if let Ok(mut textures) = self.textures.try_write() {
+            return textures.drain(..).collect::<Vec<_>>();
```
inodentry, line 39: uhm ... why? why copy all the stuff from one Vec into a new Vec to return? This can be much more efficiently done by taking the Vec out and leaving a default empty Vec in its place:
```suggestion
            return std::mem::take(&mut *textures);
```

StarArawn: I typically shy away from `std::mem` but in this case I think it makes sense. 👍

StarArawn:
Fixes: #277

src/render/extract.rs
```diff
@@ -74,6 +76,7 @@ pub(crate) struct ExtractedTilemapTexture {
     pub texture_size: TilemapTextureSize,
     pub spacing: TilemapSpacing,
     pub texture: TilemapTexture,
+    pub filtering: FilterMode,
```
bzm3r, line 79: Minor nit, maybe this should be called `filter` or `filter_mode`?

bzm3r (approved):
Looks good to me overall. Just one minor nit.

Issue #282, opened 2022-09-23
Add AABB culling for rendering chunks.


StarArawn:
We can speed up performance by not rendering rendering chunks that are off screen. This can be done by using bevy's existing camera frustum and an AABB stored on the `RenderChunk`.

Visibility check would go here:
https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/src/render/prepare.rs#L178

Use this function to check intersection with AABB:
https://github.com/bevyengine/bevy/blob/dc3b4b6c850898c922dff9fd6d312823e07096f1/crates/bevy_render/src/primitives/mod.rs#L176

AABB is in 3D space but we can likely just use 0 and 1 for the min and max respectively.

Verification can be done by either printing out draw calls or using render doc to see if render chunks off screen are still rendering.

PR #288, opened 2022-09-25
AABB-based frustum culling of `RenderChunk2d`.

bzm3r:
Fixes: https://github.com/StarArawn/bevy_ecs_tilemap/issues/282

### DIFF
--- a/src/helpers/transform.rs
+++ b/src/helpers/transform.rs
@@ -1,12 +1,14 @@
 use crate::tiles::TilePos;
-use crate::{TilemapGridSize, TilemapType};
-use bevy::math::{UVec2, Vec2};
+use crate::{TilemapGridSize, TilemapTileSize, TilemapType};
+use bevy::math::{UVec2, Vec2, Vec3, Vec3A};
+use bevy::prelude::Transform;
+use bevy::render::primitives::Aabb;
 
 /// Calculates the world-space position of the bottom-left of the specified chunk.
 pub fn chunk_index_to_world_space(
     chunk_index: UVec2,
     chunk_size: UVec2,
-    grid_size: Vec2,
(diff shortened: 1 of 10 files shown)


src/helpers/transform.rs
```diff
@@ -1,19 +1,67 @@
...
+    let minimum = Vec2::new(0.0, 0.0) - delta;
+    let maximum =
+        chunk_index_to_world_space(UVec2::new(1, 1), chunk_size, grid_size, map_type) + delta;
+    Aabb::from_min_max(minimum.extend(0.0), maximum.extend(1.0))
+}
+
+/// Apply a [`Transform`] to an [`Aabb`].
+///
+/// Workaround for: https://github.com/bevyengine/bevy/issues/6090
+pub fn apply_transform_to_aabb(transform: Transform, aabb: Aabb) -> Aabb {
```
bzm3r, line 46: Not sure if this is right.

bzm3r: This is unnecessary, it turns out.

StarArawn (changes requested):
Overall things are looking really good here!

It looks like frustum culling is broken for diamond isometric maps. From what I can tell it's caused by an incorrect AABB size. I would also like for us to have a setting to turn on/off the frustum culling as some people might want to disable that feature. Maybe as a component on the tilemap entity?

Affinator:
> 

What about a setting in the RenderSettings? This will not be decided per tilemap, but globally...

bzm3r:
@Affinator I think having the flexibility of making it a tilemap specific setting is good? I have set the default to be "Frustum culling on", so that it 99% of the time you don't have to think about it.

StarArawn (approved):

PR #293, opened 2022-09-27
Incremental improvements to hex functionality

bzm3r:
Some fixes to hex neighbor functions turned into general improvements based on the recent helpers rework.

Depends on: #291

### DIFF
--- a/examples/neighbors.rs
+++ b/examples/neighbors.rs
@@ -6,10 +6,8 @@ mod helpers;
 use crate::KeyCode::{Key0, Key1, Key2, Key3, Key4, Key5};
 use helpers::camera::movement as camera_movement;
 
-// Press SPACE to change map type. Hover over mouse tiles to highlight their labels.
-//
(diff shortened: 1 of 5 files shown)


examples/neighbors.rs
```diff
@@ -0,0 +1,459 @@
...
+                    if let Ok(mut tile_text) = tile_label_q.get_mut(tile_entity) {
+                        for mut section in tile_text.sections.iter_mut() {
+                            section.style.color = Color::BLUE;
+                        }
+                        commands.entity(tile_entity).insert(NeighborHighlight);
+                    }
+                }
+            }
+
+            let selected_hex_direction = if keyboard_input.pressed(Key0) {
```
StarArawn, line 385: Can we use `KeyCode::Key0` here? I think it's just a little clearer that way.

StarArawn (approved):
Just one stylistic change, but you can ignore.

PR #296, opened 2022-09-28
Generating hexagonal hex maps.

bzm3r:
We now have the technology!

Depends on #293

### DIFF
--- a/examples/hexagon_generation.rs
+++ b/examples/hexagon_generation.rs
@@ -223,9 +223,7 @@ fn main() {
         .insert_resource(WindowDescriptor {
             width: 1270.0,
             height: 720.0,
-            title: String::from(
-                "Hexagon Neighbors - Hover over a tile, and then press 0-5 to mark neighbors",
-            ),
+            title: String::from("Generating a hexagonal hex map"),
             ..Default::default()
         })
         .insert_resource(ImageSettings::default_nearest())
--- a/examples/hexagon_generation.rs
+++ b/examples/hexagon_generation.rs
@@ -1,0 +1,240 @@
+use bevy::{prelude::*, render::texture::ImageSettings};
+use bevy_ecs_tilemap::prelude::*;
+mod helpers;
+use helpers::camera::movement as camera_movement;
+
+// Press SPACE to change map type. Hover over a tile to highlight its label (red) and those of its
+// neighbors (blue). Press and hold one of keys 0-5 to mark the neighbor in that direction (green).
+
+// You can increase the MAP_SIDE_LENGTH, in order to test larger maps but just make sure that you run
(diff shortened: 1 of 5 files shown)


src/helpers/filling.rs
```diff
@@ -97,3 +100,78 @@ pub fn fill_tilemap_rect_color(
...
+    hexagon
+}
+
+/// Fills a hexagonal region with the given `tile_texture`.
+///
+/// The rectangular region is defined by an `origin` in [`TilePos`](crate::tiles::TilePos), and a
+/// `radius`.
+///
+/// Tiles that do not fit in the tilemap will not be created.
+pub fn fill_tilemap_hexagon(
```
StarArawn, line 148: It might be worth mentioning the cost here? This seems like O(n^2) to me.

examples/hexagon_generation.rs
```diff
@@ -0,0 +1,240 @@
...
+    }
+}
+
+fn main() {
+    App::new()
+        .insert_resource(WindowDescriptor {
+            width: 1270.0,
+            height: 720.0,
+            title: String::from(
+                "Hexagon Neighbors - Hover over a tile, and then press 0-5 to mark neighbors",
```
StarArawn, line 227: I think the title needs to change here. :)

StarArawn (changes requested):

src/helpers/filling.rs
```diff
@@ -97,3 +100,78 @@ pub fn fill_tilemap_rect_color(
...
+    let tile_positions = generate_hexagon(
+        AxialPos::from_tile_pos_given_coord_system(&origin, hex_coord_system),
+        radius,
+    )
+    .into_iter()
+    .map(|axial_pos| axial_pos.as_tile_pos_given_coord_system(hex_coord_system))
+    .collect::<Vec<TilePos>>();
+
+    for tile_pos in tile_positions {
+        let tile_entity = commands
```
alice-i-cecile, line 166: This will be measurably faster with spawn_batch. Probably want to change the sibling to match? Or just open an issue over it.

src/helpers/filling.rs
alice-i-cecile: Yeah, this is pretty inescapably (and obviously) n^2. I don't think it should be called out.

src/helpers/filling.rs
bzm3r: See: https://github.com/StarArawn/bevy_ecs_tilemap/issues/300

src/helpers/filling.rs
StarArawn: > Yeah, this is pretty inescapably (and obviously) n^2. I don't think it should be called out.
agreed

StarArawn (approved):

PR #299, opened 2022-09-30
Allow user to provide a vector of image handles, each to be used as a tile texture. 

bzm3r:
All must be of the same size. 

Depends on #296 because of example.


src/map/mod.rs
```diff
@@ -70,9 +73,86 @@ impl From<UVec2> for TilemapSize {
...
+    /// Each tile's texture has its own image asset (each asset must have the same size), so there
+    /// is a vector of image assets.
+    ///
+    /// Each image should have the same size, identical to the provided `TilemapTileSize`. If this
+    /// is not the case, a panic will be thrown during the verification when images are being
+    /// extracted to the render world.
+    ///
+    /// This only makes sense to use when the `"atlas"` feature is NOT enabled, as texture arrays
+    /// are required to handle storing an array of textures. Therefore, this variant is only
+    /// available when `"atlas"` is not enabled.
```
StarArawn, line 89: This is pretty much exactly what I had in mind. 👍

PR #301, opened 2022-10-03
fix spacing being added to atlas size

antoinePaulinB7:
If you want the proper count of the number of tiles / columns, the atlas size should be taken as is.


antoinePaulinB7:
@StarArawn I noticed that you mentioned this bit of code is contentious, so I thought what better to do than jump into as well.

bzm3r:
@antoinePaulinB7 Thanks for this PR! We just merged in some changes that are causing a conflict with your branch. Would you mind merging `main` back into your branch, to resolve the conflict?

antoinePaulinB7:
@bzm3r done :)

src/render/extract.rs
```diff
@@ -96,10 +96,8 @@ impl ExtractedTilemapTexture {
                     it is being extracted as a texture!",
                 );
                 let texture_size: TilemapTextureSize = image.size().into();
-                let tile_count_x =
-                    (texture_size.x + tile_spacing.x) / (tile_size.x + tile_spacing.x);
-                let tile_count_y =
-                    (texture_size.y + tile_spacing.y) / (tile_size.y + tile_spacing.y);
+                let tile_count_x = (texture_size.x) / (tile_size.x + tile_spacing.x);
```
bzm3r, line 99: Is this making the assumption that every tile, including the last one in the atlas will have padding? Is that the right assumption here?

bzm3r (changes requested):
This makes sense to me at first glance, but I really think we need to put together a proper test for it. @antoinePaulinB7 Can you put in an asset and an example that clearly shows why this works, and the old method would not?

src/render/extract.rs
antoinePaulinB7: That is indeed the assumption.

If we wanted to get rid of this assumption, I propose that we take inspiration from bevy's own bevy_sprite `from_grid_with_padding`.

antoinePaulinB7:
@bzm3r Hopefully, I'll have time to put together a test before the end of the week.

src/render/extract.rs
bzm3r: @antoinePaulinB7 I have been thinking about it, and I think it's a fine assumption, because it provides a lot of regularity and eliminates special cases. Perhaps it is something we can document, so users know that this is what we expect. I will create a separate issue for it if this PR gets merged, to work on documenting that.

bzm3r:
@StarArawn Can you also review this PR, and see if it makes sense to you? It looks good to me!

Issue #260, opened 2022-09-06
WGPU Validation Error (CommandEncoder::copy_texture_to_texture)


aleokdev:
Using [this map](https://github.com/StarArawn/bevy_ecs_tilemap/files/9495689/map.zip) with the LDTK example results in a WGPU validation error:
```
thread 'Compute Task Pool (1)' panicked at 'wgpu error: Validation Error

Caused by:
    In CommandEncoder::copy_texture_to_texture
    Copy error
    copy of X 192..208 would end up overrunning the bounds of the Source texture of X size 203

', /home/aleok/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.13.1/src/backend/direct.rs:2391:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/panicking.rs:142:14
   2: wgpu::backend::direct::default_error_handler
             at /home/aleok/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.13.1/src/backend/direct.rs:2391:5
   3: core::ops::function::Fn::call
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/ops/function.rs:77:5
   4: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/boxed.rs:1965:9
   5: wgpu::backend::direct::ErrorSinkRaw::handle_error
             at /home/aleok/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.13.1/src/backend/direct.rs:2377:17
   6: wgpu::backend::direct::Context::handle_error
             at /home/aleok/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.13.1/src/backend/direct.rs:261:9
   7: wgpu::backend::direct::Context::handle_error_nolabel
             at /home/aleok/.cargo/registry/src/github.com-1ecc6299db9ec823/wgpu-0.13.1/src/backend/direct.rs:273:9
[...]

PR #309, opened 2022-10-09
Spacing fixes

StarArawn:
Fixes: #260
Closes: #301


StarArawn:
@antoinePaulinB7 
I kept your commits here, fixed some of the math, and provided an example.

src/render/extract.rs
```diff
@@ -96,10 +96,8 @@ impl ExtractedTilemapTexture {
                     it is being extracted as a texture!",
                 );
                 let texture_size: TilemapTextureSize = image.size().into();
-                let tile_count_x =
-                    (texture_size.x + tile_spacing.x) / (tile_size.x + tile_spacing.x);
-                let tile_count_y =
-                    (texture_size.y + tile_spacing.y) / (tile_size.y + tile_spacing.y);
+                let tile_count_x = ((texture_size.x) / (tile_size.x + tile_spacing.x)).floor();
```
StarArawn, line 99: You need to floor here because otherwise you can get tile counts that are larger than what's available in the atlas.

src/render/texture_array_cache.rs
```diff
@@ -206,12 +204,15 @@ impl TextureArrayCache {
...
-                            (i as f32 / columns).floor() * (tile_size.y + spacing.y) as f32;
+                        let columns = ((texture_size.x as f32)
+                            / (tile_size.x as f32 + spacing.x as f32))
+                            .floor();
+                        let sprite_sheet_x: f32 = (i as f32 % columns).floor()
+                            * (tile_size.x + spacing.x) as f32
+                            + spacing.x;
+                        let sprite_sheet_y: f32 = (i as f32 / columns).floor()
+                            * (tile_size.y + spacing.y) as f32
+                            + spacing.y;
```
StarArawn, line 215: Don't divide by 2 here. This is because spacing is treated as the "space" between tiles. This is what tiled map editor does and I think it's something we should stick with.

examples/spacing.rs
```diff
@@ -0,0 +1,80 @@
+use bevy::{prelude::*, render::texture::ImageSettings};
```
StarArawn, line 1: Simple example based off of the layer example(just to show multiple tiles). The spacing is 8 pixels between each tile on both the X and Y axes.

bzm3r (approved):
Yes, this looks good to me!

bzm3r:
@antoinePaulinB7 This is similar to your PR, but the main change is that the spacing is to the left/below tiles, rather than all around them.

PR #312, opened 2022-10-10
Add compat info about bevy-track and bevy main

geirsagberg:


bzm3r (approved):
This looks good to me overall! I left a suggestion, but it's not one we have to use.

Issue #314, opened 2022-10-11
Compile errors when `render` feature is disabled


Seldom-SE:
`bevy_ecs_tilemap` doesn't compile when I disable the `render` feature:

```toml
[dependencies]
bevy = "0.8"
bevy_ecs_tilemap = { version = "0.8", default-features = false }
```

```
   Compiling bevy_ecs_tilemap v0.8.0
error[E0432]: unresolved import `crate::render`
 --> C:\Users\Hayde\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_ecs_tilemap-0.8.0\src\array_texture_preload.rs:3:5
  |
3 |     render::TextureArrayCache,
  |     ^^^^^^
  |     |
  |     unresolved import
  |     help: a similar path exists: `bevy::render`

error[E0277]: the trait bound `for<'r, 's, 't0, 't1, 't2, 't3, 't4, 't5, 't6, 't7> fn(bevy::render::Extract<'r, 's, Res<'t0, Assets<bevy::prelude::Image>>>, bevy::render::Extract<'t1, 't2, Res<'t3, ImageSettings>>, bevy::render::Extract<'t4, 't5, Res<'t6, ArrayTextureLoader>>, ResMut<'t7, [type error]>) {extract}: IntoSystem<(), (), _>` is not satisfied
   --> C:\Users\Hayde\.cargo\registry\src\github.com-1ecc6299db9ec823\bevy_ecs_tilemap-0.8.0\src\lib.rs:58:66
    |
58  |             render_app.add_system_to_stage(RenderStage::Extract, array_texture_preload::extract);
    |                        -------------------                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `IntoSystem<(), (), _>` is not implemented for `for<'r, 's, 't0, 't1, 't2, 't3, 't4, 't5, 't6, 't7> fn(bevy::render::Extract<'r, 's, Res<'t0, Assets<bevy::prelude::Image>>>, bevy::render::Extract<'t1, 't2, Res<'t3, ImageSettings>>, bevy::render::Extract<'t4, 't5, Res<'t6, ArrayTextureLoader>>, ResMut<'t7, [type error]>) {extract}`
    |                        |
    |                        required by a bound introduced by this call
    |
    = help: the following other types implement trait `IntoSystemDescriptor<Params>`:
              ExclusiveSystemCoerced
              ExclusiveSystemDescriptor
[...]

PR #317, opened 2022-10-14
Simple bug fixes

### LINKED ISSUE #315: TileStorage checked_remove does not actually check.
Like the title says. This should be a quick fix.

Issue in code:
https://github.com/StarArawn/bevy_ecs_tilemap/blob/de8d005331dde7d9ebb8c4fe1b36769b84e44be5/src/tiles/storage.rs#L84

bzm3r:
Fixes #315 and #314

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -51,15 +51,11 @@ impl Plugin for TilemapPlugin {
 
         app.add_system_to_stage(CoreStage::First, update_changed_tile_positions);
 
-        #[cfg(not(feature = "atlas"))]
+        #[cfg(all(not(feature = "atlas"), feature = "render"))]
         {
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -54,8 +54,12 @@ impl Plugin for TilemapPlugin {
         #[cfg(not(feature = "atlas"))]
```
bzm3r, line 54: Should I just make this `#[cfg(all(not(feature = "atlas"), feature = "render"))]`?

StarArawn: Yes

StarArawn (changes requested):

StarArawn (approved):

PR #327, opened 2022-10-26
Implement Default for Neighbors

st0rmbtw:


bzm3r:
This will be addressed by #328 too!

StarArawn:
Addressed in #328

PR #329, opened 2022-10-27
Disable tests in CI

bzm3r:
Commented it out, so we can re-introduce it later, once we actually have tests.

The benefit: for now, we get faster CI runs.


alice-i-cecile (approved):
🥲

PR #330, opened 2022-10-27
Update bevy-track branch to newer bevy commit

rparrett:
Please feel free to close if this isn't helpful or use the code however you wish.

This is mostly just tedious example work, but replacing `ImageSettings` was somewhat interesting so I'd look closely at that part.


bzm3r (approved):

bzm3r:
@rparrett 🎉

PR #338, opened 2022-11-09
Revert change to how web examples should be run

bzm3r:


StarArawn (approved):
Looks good! I also tested this flow and it works fine.

PR #343, opened 2022-11-15
Tile maps with an even number of tiles need a half-tile offset applied.

### LINKED ISSUE #342: get_tilemap_center_transform is broken for some tilemaps
### Version tested

Latest commit, 0.9.0

### Problem

Tiles appear to be half a tile too far left and down.

image

### Repro

For dramatic effect, modify `basic.rs` with
```rust
WindowDescriptor {
    width: 540.0,
    height: 540.0,
    /* ... */
}
```
Or just resize the window at runtime.

`cargo run --example basic`

bzm3r:
Fixes #342

### DIFF
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -101,8 +101,8 @@ fn main() {
     App::new()
         .add_plugins(DefaultPlugins.set(WindowPlugin{
             window: WindowDescriptor {
-                width: 1270.0,
-                height: 720.0,
+                width: 540.0,
+                height: 540.0,
                 title: String::from(
                     "Basic Example - Press Space to change Texture and H to show/hide tilemap.",
                 ),
--- a/examples/basic.rs
+++ b/examples/basic.rs
@@ -3,6 +3,9 @@ use bevy_ecs_tilemap::prelude::*;
 
 mod helpers;
 
+pub const MAP_SIZE: TilemapSize = TilemapSize { x: 32, y: 32 };
+pub const TILE_SIZE: TilemapTileSize = TilemapTileSize { x: 16.0, y: 16.0 };
+
 fn startup(
     mut commands: Commands,
     asset_server: Res<AssetServer>,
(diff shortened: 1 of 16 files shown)


rparrett:
I am not sure if this is working properly for hex and iso maps.

image
image

bzm3r:
@rparrett I just pushed a commit that now calculates the "center" based on the corner points of the central-ish tile. Can you give it a whirl for hexagonal/isometric tiles, and let me know how it looks now?

If it looks good, great. If not then, this is getting finicky enough that it's not a great use of time to solve it...so maybe we just say "eh, that's to be expected for these non-regular tile types?". 

As part of this commit, I have included updates to the helpers which calculate corner positions. So, maybe for more specific cases, users can take advantage of that to figure out how they want to center their tiles? Not sure.

rparrett:
I am not personally invested in hex / iso maps and this does seem like quite a lot to me.

I tried an approach like this and it seems to work well enough with hex/rect tilemaps with odd and even tile counts.

```rust
/// Calculates a [`Transform`] for a tilemap that places it so that the center tile is at
/// `(0.0, 0.0, 0.0)` in world space.
pub fn get_tilemap_center_transform(
    size: &TilemapSize,
    grid_size: &TilemapGridSize,
    map_type: &TilemapType,
    z: f32,
) -> Transform {
    let low = TilePos::new(0, 0).center_in_world(grid_size, map_type);
    let high = TilePos::new(size.x - 1, size.y - 1).center_in_world(grid_size, map_type);

    let diff = high - low;

    Transform::from_xyz(-diff.x / 2., -diff.y / 2., z)
}
```

Maybe it doesn't work with iso. I'm not immediately sure how to modify that example to test and I'm out of time for now.

bzm3r:
@rparrett Yeah, that seems way better, to be honest. I'm facepalming, not sure why I got distracted with calculating corner positions.

rparrett:
The new `get_tilemap_center_transform` worked well for me when I tested it in the absence of the rest of the changes to `center_world_in_pos`, etc.

But in the current changeset, hexagon centering seems broken again.

There are also some unrelated changes in the changeset that snuck in like a (bug fix?) in `filling.rs` and changes to examples.

bzm3r:
@rparrett Yeah, I am fine with putting those in as part of the current change set, as I didn't know how to separate it out, and they were small enough that I would forget them otherwise.

PR #356, opened 2022-11-24
Examples cleanup

rparrett:
I feel like there is a lot left to do here, but this branch is getting large and I'm running out of time.

I tested all examples and their keyboard functionality. `texture_vec` `texture_container` were already broken and the problem wasn't obvious so I didn't touch them.

My main goal was to center a bunch of tilemaps and improve consistency of tilemap spawning code between examples, but I stumbled on a few other little cleanups.

I'd recommend reviewing individual commits.


examples/helpers/ldtk.rs
```diff
@@ -25,7 +25,7 @@ impl Plugin for LdtkPlugin {
 }
 
 #[derive(TypeUuid)]
-#[uuid = "e51081d0-6168-4881-a1c6-4249b2000d7f"]
+#[uuid = "abf9eaf2-f21c-4b46-89b0-8aa5c42199af"]
```
StarArawn, line 28: Was this sharing a UUID from somewhere else? 🤔

rparrett: Just with the Tiled example's `TiledMap`. Not the most likely collision, but I wouldn't put it totally outside the realm of possibility.

examples/hex_neighbors.rs
```diff
@@ -96,23 +99,23 @@ fn spawn_tile_labels(
...
-    for (map_transform, map_type, grid_size, tilemap_storage) in tilemap_q.iter() {
+    for (map_type, grid_size, tilemap_storage) in tilemap_q.iter() {
         for tile_entity in tilemap_storage.iter().flatten() {
             let tile_pos = tile_q.get(*tile_entity).unwrap();
             let tile_center = tile_pos.center_in_world(grid_size, map_type).extend(1.0);
-            let transform = *map_transform * Transform::from_translation(tile_center);
-            commands
-                .entity(*tile_entity)
-                .insert(Text2dBundle {
+            let transform = Transform::from_translation(tile_center);
```
bzm3r, line 106: Does this work because the tilemap is being centered now? 

Regardless, I am not sure if map_transform should be removed, because it is worth knowing this should be done in an example.

For example, what happens if a map is panned? Does the example still work?

examples/hex_neighbors.rs
```diff
@@ -210,9 +221,11 @@ fn swap_map_type(
                 _ => unreachable!(),
             }
 
+            *map_transform = get_tilemap_center_transform(&map_size, &grid_size, &map_type, 0.0);
+
             for (tile_pos, mut tile_label_transform) in tile_label_q.iter_mut() {
                 let tile_center = tile_pos.center_in_world(&grid_size, &map_type).extend(1.0);
-                *tile_label_transform = *map_transform * Transform::from_translation(tile_center);
+                *tile_label_transform = Transform::from_translation(tile_center);
```
bzm3r, line 228: Again, does this still work if the map is panned off center?

examples/hex_neighbors.rs
```diff
@@ -447,24 +465,18 @@ fn main() {
...
-        .add_system_to_stage(
-            CoreStage::Update,
-            hover_highlight_tile_label.after(swap_map_type),
-        )
-        .add_system_to_stage(
-            CoreStage::Update,
-            highlight_neighbor_label.after(hover_highlight_tile_label),
-        )
+        .add_system(swap_map_type)
+        .add_system(hover_highlight_tile_label.after(swap_map_type))
```
bzm3r, line 479: This is awesome, we should really move away from using stages!

examples/mouse_to_tile.rs
```diff
@@ -39,52 +39,70 @@ pub struct TileHandleIso(Handle<Image>);
...
-    let tile_handle_hex_col: Handle<Image> = asset_server.load("bw-tile-hex-col.png");
-    let tile_handle_square: Handle<Image> = asset_server.load("bw-tile-square.png");
-    let font: Handle<Font> = asset_server.load("fonts/FiraSans-Bold.ttf");
-
-    commands.insert_resource(TileHandleIso(tile_handle_iso));
-    commands.insert_resource(TileHandleHexCol(tile_handle_hex_col));
-    commands.insert_resource(TileHandleHexRow(tile_handle_hex_row));
-    commands.insert_resource(TileHandleSquare(tile_handle_square));
-    commands.insert_resource(FontHandle(font));
+impl FromWorld for TileHandleHexCol {
```
bzm3r, line 42: Very cool!

examples/mouse_to_tile.rs
```diff
@@ -238,9 +265,11 @@ fn swap_map_type(
                 }
             }
 
+            *map_transform = get_tilemap_center_transform(&map_size, &grid_size, &map_type, 0.0);
+
             for (tile_pos, mut tile_label_transform) in tile_label_q.iter_mut() {
                 let tile_center = tile_pos.center_in_world(&grid_size, &map_type).extend(1.0);
-                *tile_label_transform = *map_transform * Transform::from_translation(tile_center);
+                *tile_label_transform = Transform::from_translation(tile_center);
```
bzm3r, line 272: See other comments: we need to check to make sure that this still works when the map is panned

bzm3r (changes requested):
This is a welcome standardization fix. Thanks for taking the time to do it.

My main question is around the removal of map_transform in the multiplication with the tile center's transform. Can you comment?

examples/hex_neighbors.rs
rparrett: Fascinating question. The examples work just as they did before when the map is panned and when the map type is changed.

Although now that I am looking at this again, it does seem like the map transform *should* be involved here. (Despite `TilePos::center_in_world`'s claims about returning a world position, I don't think it is?)

I'll take a closer look at this in a bit.

examples/hex_neighbors.rs
rparrett: I would love to continue cleaning that up later. I'm pretty sure these examples don't need any `add_*_to_stage` to work.

examples/hex_neighbors.rs
rparrett: Okay, I did some investigating.

The `Text2dBundle`'s components (including GlobalTransform) are being added directly to the Tile, and the Tiles are all children of the Tilemap.

So the text's transform is relative to the Tilemap because transform propagation is happening. This all worked out fine when the Tilemap was at 0,0 prior to this PR. But now that the Tilemap is centered, this double-adding the map transform produces the wrong result.

I'm not sure what the "bevy_ecs_tilemap" way to fix this is. To me, it looks like the text should be spawned as a separate entity, and a component like `TileTextEntity(Entity)` should be inserted onto the Tile instead. It's my understanding that "bevy_ecs_tilemap" is generally avoiding any transform or visibility propagation on purpose for performance reasos.

If that solution sounds acceptable, I'd be happy to implement it.

rparrett:
Note: there is currently (perhaps a) system ordering issue in `hexagon_generation`. On some runs, switching map types results in empty tiles in the maps. On other runs, it works fine. This behavior was also present prior to this PR, but I'm looking into it.

examples/hex_neighbors.rs
rparrett: This is what that changeset would look like for one of the examples: <https://github.com/rparrett/bevy_ecs_tilemap/commit/1b83aeee0f6ea10295a395ad91febd85ff492242>

examples/hex_neighbors.rs
bzm3r: Yeah, quite possibly. I think the only time they are definitely needed is when we have commands that need to be flushed before the next stage. I can't remember how much that's true for many of the examples.

examples/hex_neighbors.rs
bzm3r: Thanks for looking into this. 

IIUC, I think the correct answer here then is to remove `map_transform` being multiplied here, and trust that the text entity will be properly managed by bevy. 

Why? Because text is not something deeply a part of bevy_ecs_tilemap; it's just a debugging tool for the example. Likely, if players need more complex text solutions, they will have to roll their own. For now, for the sake of the example, then, there's no need to multiply by the map transform?

@StarArawn Let me know what you think.

bzm3r: @rparrett Oh, I just had a look at the example you changed. So, does wrapping the `Text2dBundle` in its own component remove the transform propagation?

rparrett: I think you understand correctly (if I do.) The map_transform multiplication prior to this PR is a bug that wasn't revealed until the map was moved.

Adding the `Text2dBundle` outside of bevy's hierarchy and into "our own hierarchy" (storing a reference to the Entity on a component) prevents the transform propagation from happening if we want that behavior to work. (At the expense of complicating all the interactions with the text)

Displaying coordinates or other sprites or whatever on top of tiles is a common thing for users to want to do, so IMHO, we should use whatever pattern we want users to use.

I am an extremely casual user of `bevy_ecs_tilemap` that hasn't followed recent developments closely, so I don't have a very good intuition for what that is.

StarArawn:
@rparrett Is this ready for review again? Also the conflicts need to be fixed.

examples/hex_neighbors.rs
bzm3r: @rparrett Okay, after talking about this with StarToaster, and thinking about it further myself, I think:

1) for now, to get the examples working, we can do what you suggested here: https://github.com/StarArawn/bevy_ecs_tilemap/pull/356#discussion_r1033070351

2) for future work, we should make an issue for TileText that follows your more general suggestion

rparrett:
Rebased. Looks like there's consensus on removing tile labels from the hierarchy. I should be able to get that done later todayl

bzm3r (approved):

PR #361, opened 2022-11-28
Remove unused section from Cargo.toml

rparrett:
Fixes this warning seen when running any example
```
warning: unused manifest key: target.wasm32-unknown-unknown.runner
```

As noted here: https://github.com/jakobhellermann/wasm-server-runner#step-2 this only works in an individual's cargo config.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,9 +62,6 @@ features = [
     "bevy_sprite",
 ]
 
-[target.wasm32-unknown-unknown]
-runner = "wasm-server-runner"
-
 [[example]]
 name = "ldtk"
 path = "examples/ldtk/ldtk_usage.rs"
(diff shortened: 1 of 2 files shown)


bzm3r:
Big if true! Has this been tested with: https://github.com/StarArawn/bevy_ecs_tilemap#running-examples-on-web

rparrett:
Those instructions do continue to work. The removed section is completely superfluous / doing nothing for us.

However, `wasm-server-runner` really shines when configured as described in the link above. If configured properly, the seconds command isn't necessary at all. Cargo will use the configured runner automatically and all you have to do is (note: `run` instead of `build`)

```
cargo run --target wasm32-unknown-unknown --example animation --release --features atlas
```

I'd suggest changing the instructions to
```
cargo install wasm-server-runner
cargo run --target wasm32-unknown-unknown --example animation --release --features atlas
```

StarArawn (approved):

PR #363, opened 2022-11-29
Fix broken rendering of tilemaps with padding in atlas mode

### LINKED ISSUE #352: TilemapSpacing is broken when using atlas feature
### Version tested

MacOS (M1)
main branch

### Problem

However, it seems that when using the atlas texture renderer, both tilemaps look very broken, but in different ways.

`cargo run --example spacing --features=atlas`
atlas

rparrett:
Fixes #352 

`var columns` was mistakenly adding rather than subtracting one side of the spacing, potentially causing the wrong tiles to be displayed. This bug was not visible in the spacing example. However, this would have produced the wrong result when spacing was greater than or equal to tile size.

`sprite_sheet_x` and `sprite_sheet_y` were not accounting for the what bevy calls the "offset" -- the spacing in the left gutter of the texture.

Before:
`cargo run --example spacing --features=atlas`
image

After
`cargo run --example spacing --features=atlas`
image

Note: the spacing example is intentionally displaying an incorrectly configured tilemap in blue.

### DIFF
--- a/src/render/shaders/tilemap_vertex.wgsl
+++ b/src/render/shaders/tilemap_vertex.wgsl
@@ -57,12 +57,11 @@ fn vertex(vertex_input: VertexInput) -> VertexOutput {
 
     var texture_index: u32 = u32(current_animation_frame);
 
+    #ifdef ATLAS
     var columns: u32 = u32((tilemap_data.texture_size.x - tilemap_data.spacing.x) / (tilemap_data.tile_size.x + tilemap_data.spacing.x));
-
     var sprite_sheet_x: f32 = tilemap_data.spacing.x + floor(f32(texture_index % columns)) * (tilemap_data.tile_size.x + tilemap_data.spacing.x);
     var sprite_sheet_y: f32 = tilemap_data.spacing.y + floor(f32(texture_index / columns)) * (tilemap_data.tile_size.y + tilemap_data.spacing.y);
 
-    #ifdef ATLAS
     var start_u: f32 = sprite_sheet_x / tilemap_data.texture_size.x;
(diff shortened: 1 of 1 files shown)


bzm3r:
@rparrett thanks for your patience on this; I've asked StarArawn to review, but he is currently super busy, so it may take a while!

But this does look good to me!

src/render/shaders/tilemap_vertex.wgsl
```diff
@@ -57,12 +57,11 @@ fn vertex(vertex_input: VertexInput) -> VertexOutput {
 
     var texture_index: u32 = u32(current_animation_frame);
 
-    var columns: u32 = u32((tilemap_data.texture_size.x + tilemap_data.spacing.x) / (tilemap_data.tile_size.x + tilemap_data.spacing.x));
-
-    var sprite_sheet_x: f32 = floor(f32(texture_index % columns)) * (tilemap_data.tile_size.x + tilemap_data.spacing.x);
-    var sprite_sheet_y: f32 = floor(f32(texture_index / columns)) * (tilemap_data.tile_size.y + tilemap_data.spacing.y);
-
     #ifdef ATLAS
+    var columns: u32 = u32((tilemap_data.texture_size.x - tilemap_data.spacing.x) / (tilemap_data.tile_size.x + tilemap_data.spacing.x));
```
bzm3r, line 61: I think this makes sense to me? But I'm wondering if we can add some more comments here, similar to what the PR text includes, for future readers?

bzm3r (changes requested):

StarArawn (approved):

StarArawn (approved):

PR #367, opened 2022-12-18
Fix incorrect tile UVs when using atlas feature on AMD hardware

rparrett:
Fixes an issue reported by @odecay on discord.

This is specific to AMD hardware and the issue and fix were reproduced by @odecay and @bzm3r. I am unable to reproduce on any of my non-AMD HW.

Here's a changeset on their fork that reproduces the issue:
https://github.com/StarArawn/bevy_ecs_tilemap/compare/main...odecay:bevy_ecs_tilemap:main

That example should show something like this:
image
But shows this instead:
image

It seems that there's some sort of floating point imprecision thing going on where `var columns` ends up as `2` rather than the correct `3` after being cast as `u32` / implicitly `floor`ed. This messes up the rest of the UV calculations.

It is possible to induce this same visual result on hardware that doesn't exhibit this bug by subtracting 1 from `var columns`.

### DIFF
--- a/src/render/shaders/tilemap_vertex.wgsl
+++ b/src/render/shaders/tilemap_vertex.wgsl
@@ -60,7 +60,7 @@ fn vertex(vertex_input: VertexInput) -> VertexOutput {
     #ifdef ATLAS
     // Get the top-left corner of the current frame in the texture, accounting for padding around the whole texture
     // as well as spacing between the tiles.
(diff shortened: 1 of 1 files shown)


bzm3r:
Big if true! And it is true!

bzm3r (approved):

odecay:
Can confirm this change fixes the issue for me :+1:

odecay (approved):

PR #374, opened 2022-12-27
Add missing #[reflect(Component)] annotations

johanhelsing:
Adds #[reflect(Component)] for all components except AnimatedTile, which doesn't implement Default.


bzm3r (approved):

bzm3r:
Thank you!

Issue #375, opened 2022-12-28
Should `fill_tilemap` be adding tiles as children to the tilemap entity?


SirHall:
`fill_tilemap` causes the `bench` command to take a *very* long time to load (long enough that I don't actually know).
If we look at its definition in filling.rs:
```rust
/// Fills an entire tile storage with the given tile.
pub fn fill_tilemap(
    texture_index: TileTextureIndex,
    size: TilemapSize,
    tilemap_id: TilemapId,
    commands: &mut Commands,
    tile_storage: &mut TileStorage,
) {
    for x in 0..size.x {
        for y in 0..size.y {
            let tile_pos = TilePos { x, y };
            let tile_entity = commands
                .spawn(TileBundle {
                    position: tile_pos,
                    tilemap_id,
                    texture_index,
                    ..Default::default()
                })
                .id();
            commands.entity(tilemap_id.0).add_child(tile_entity);
            tile_storage.set(&tile_pos, tile_entity);
        }
    }
}
```

The culprit from my testing has been `commands.entity(tilemap_id.0).add_child(tile_entity);`.
Without that line the benchmark launched in a mere few seconds, not very methodical - but it works.

One option is to take the loop, pull out the `commands.entity(tilemap_id.0).add_child(tile_entity);` call, switch to using `add_children` and place the loop inside that:
```rust
/// Fills an entire tile storage with the given tile.
pub fn fill_tilemap(
    texture_index: TileTextureIndex,
    size: TilemapSize,
    tilemap_id: TilemapId,
    commands: &mut Commands,
    tile_storage: &mut TileStorage,
) {
    let mut tilemap = commands.entity(tilemap_id.0);
    tilemap.add_children(|parent| {
        for x in 0..size.x {
            for y in 0..size.y {
                let tile_pos = TilePos { x, y };
                let tile_entity = {
                    parent
                        .spawn(TileBundle {
                            position: tile_pos,
                            tilemap_id,
                            texture_index,
                            ..Default::default()
                        })
                        .id()
                };
                tile_storage.set(&tile_pos, tile_entity);
            }
        }
    });
}
```

This fixes the loading issue and now `bench` loads quickly again.
Only issue is that while perusing the other `fill*` functions I noticed that none of them add the newly spawned tiles as children to the tilemap entity.

Is this an oversight, and should we be fixing all the other filler functions, is it not needed and rather should we be taking it out of `fill_tilemap`?

[...]

PR #377, opened 2022-12-29
Invert `fill_tilemap` usage of `add_child` to the tilemap parent entity to instead use `add_children`

SirHall:
Please refer to and read through #375, decline this pull request if this is the wrong path.

### DIFF
--- a/src/helpers/filling.rs
+++ b/src/helpers/filling.rs
@@ -15,21 +15,25 @@ pub fn fill_tilemap(
     commands: &mut Commands,
     tile_storage: &mut TileStorage,
 ) {
-    for x in 0..size.x {
-        for y in 0..size.y {
-            let tile_pos = TilePos { x, y };
-            let tile_entity = commands
-                .spawn(TileBundle {
-                    position: tile_pos,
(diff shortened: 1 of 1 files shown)


bzm3r (changes requested):
This makes sense to me overall, but I you're right that we don't need to be adding the tiles as children of the tilemap in the first place. So maybe that will be the ultimate simplification?

StarArawn:
> This makes sense to me overall, but I you're right that we don't need to be adding the tiles as children of the tilemap in the first place. So maybe that will be the ultimate simplification?

I'm a little late to the party here, but we need to add tiles as children to the tilemap in order for `despawn_recursive` to work correctly or remind users that they need to despawn tiles. Otherwise this kind of thing can happen:
https://github.com/StarArawn/bevy_ecs_tilemap/pull/384
Where you have dangling tiles(tiles no longer associated with a tilemap). This results in a panic in `bevy_ecs_tilemap` right now.

SirHall:
Would you rather have all the fill methods add the tiles as children then?

Issue #320, opened 2022-10-24
Newer `tiled` dependency planned?


Danl2620:
Hi, I'm trying a custom map with the `tiled.rs` example but it throws this warning:

```
WARN bevy_asset::asset_server: encountered an error while loading an asset: Maps with external tilesets must know their file location.  See parse_with_path(Path).
```

I can't track down where that warning comes from but I'm guessing it's from the `tiled` crate dependency. That crate has made major changes to map loading since the pinned version used by `bevy_ecs_tiled` (0.9). I'm wondering if there's any plan or desire to upgrade to the latest version of `tiled` (0.10.2)?

StarArawn:
> I'm wondering if there's any plan or desire to upgrade to the latest version of tiled (0.10.2)?
No plan currently, but PR's for these types of things are always welcome. 👍

Danl2620:
Ok, sounds good. I'll take a look this weekend.

bjorn:
Duplicate of #249?

Issue #294, opened 2022-09-28
Document pros/cons of `atlas` feature vs. default array system 


bzm3r:
I enabled the `atlas` feature, and it seems to work now. However, I don't really know what the drawbacks or the major differences of enabling that feature are, since there is not really any documentation of that feature.

_Originally posted by @Floyden in https://github.com/StarArawn/bevy_ecs_tilemap/issues/289#issuecomment-1258036358_

rparrett:
From StarArawn on Discord:

> We have two rendering paths:
> 1. Array textures
> 2. Atlas textures 
> Array textures are preferred on by default. These array textures support a lot of default GPU texture features natively. Atlas by comparison does not support things like non-point filtering, mipmaps, etc. Atlas is primarily still around because it's the only thing that works currently on webGL2. I've opened an issue as array textures should work:
https://github.com/gfx-rs/wgpu/issues/3197
> 
> From a user perspective array texture does everything that atlas texture does and more. Atlases the user loads in are automatically converted into array textures. When the atlas feature is on the atlas is instead sent directly to the GPU.

PR #381, opened 2023-01-01
Upgrade tiled dep and add support for image-per-tile tilesets.

### LINKED ISSUE #249: Update `tiled` to version `0.10`
New tiled version: https://crates.io/crates/tiled

dvogel:
This does two things:

* Adds support for tilesets that use a separate image per tile.
* Upgrades the `tiled` dependency. Should be no change in actual functionality but the 0.10 version is much easier to work with than the 0.9 version. (fixes #249 and #320).

Here is a screenshot of the example running with this code:

Screenshot from 2023-01-01 11-08-39

Here is a screenshot of a map loaded with the new image-per-tile support (this file has 5 layers and each is rendered as I see it in Tiled):

Screenshot from 2023-01-01 11-12-17


examples/helpers/tiled.rs
```diff
@@ -135,115 +181,159 @@ pub fn process_loaded_maps(
...
+                                            let mut mapped_y = y;
+                                            if tiled_map.map.orientation
+                                                == tiled::Orientation::Orthogonal
+                                            {
+                                                mapped_y = (tiled_map.map.height - 1) - y;
+                                            }
 
-                                let mapped_x = x as usize;
-                                let mapped_y = mapped_y as usize;
+                                            let mapped_x = x as usize;
```
rparrett, line 243: `mapped_x` and `mapped_y` are seemingly only ever used as `i32, so the `as i32` might as well be here.

rparrett:
This doesn't appear to work with my map. It was saved with Tiled 1.4, so I think it should be compatible.

https://github.com/rparrett/taipo/blob/main/assets/textures/level1.tmx

image

rparrett:
Actually, this isn't working properly with the in-repo example either. It should look like this:
image

dvogel:
Thanks for reviewing this @rparrett. I'm not sure why I didn't notice the obvious error for the test map rendering. I'll take another pass at this soon.

dvogel:
I just pushed a change that incorporates your suggestions and fixes the test map loading. Here's screenshots of the test map being rendered as it is by default and then zoomed out to see the entire thing.

Screenshot from 2023-01-05 22-40-48
Screenshot from 2023-01-05 22-41-04

The spacing between tiles looks a little strange but I think it is a rendering artifact rather than a tile placement error because it goes away when I zoom in. Would be good to test on another machine though.

Screenshot from 2023-01-05 22-48-02

I wasn't able to test with your map @rparrett because it doesn't include the tileset image. 

Screenshot from 2023-01-05 22-46-54

examples/helpers/tiled.rs
```diff
@@ -135,115 +181,159 @@ pub fn process_loaded_maps(
...
+                                layer.layer_type()
+                            );
+                            continue;
                         };
 
-                        let grid_size = TilemapGridSize {
-                            x: tiled_map.map.tile_width as f32,
-                            y: tiled_map.map.tile_height as f32,
-                        };
+                        match tile_layer {
```
rparrett, line 213: Can we collapse this match into a let else guard also?

examples/helpers/tiled.rs
```diff
@@ -135,115 +181,159 @@ pub fn process_loaded_maps(
...
+                    for (layer_index, layer) in tiled_map.map.layers().enumerate() {
                         let offset_x = layer.offset_x;
                         let offset_y = layer.offset_y;
 
-                        let map_size = TilemapSize {
-                            x: tiled_map.map.width,
-                            y: tiled_map.map.height,
+                        let tiled::LayerType::TileLayer(tile_layer) = layer.layer_type() else {
+                            log::info!(
+                                "Skipping layer because {:?} is not supported.",
```
rparrett, line 207: It seems like the debug impl for `LayerType` prints **all** the map data somehow, making this spew several screens of spam for my map.

Maybe something like
```rust
log::info!(
    "Skipping layer {} because it is an unsupported type",
    layer.id()
);
```
would be adequate?

examples/helpers/tiled.rs
```diff
@@ -135,115 +181,159 @@ pub fn process_loaded_maps(
...
+                                    map_type,
+                                    ..Default::default()
+                                });
+
+                                layer_storage
+                                    .storage
+                                    .insert(layer_index as u32, layer_entity);
+                            }
+                            _ => {
+                                log::info!(
```
rparrett, line 331: Same thing here about the debug print.

rparrett (commented):
Looking good on my end now. (tileset image for `level1.tmx` is in the same repo)

I'd still like to take a slightly closer look when I have a few more minutes, especially with a map with multiple "single image" tilesets and layers.

I wouldn't expect the visual artifacts you mentioned to be related to this PR.

examples/helpers/tiled.rs
```diff
@@ -135,115 +181,159 @@ pub fn process_loaded_maps(
...
-                                        flip: TileFlip {
-                                            x: map_tile.flip_h,
-                                            y: map_tile.flip_v,
-                                            d: map_tile.flip_d,
-                                        },
-                                        ..Default::default()
-                                    })
-                                    .id();
-                                tile_storage.set(&tile_pos, tile_entity);
+                                let texture = if tiled_map.tile_texture_indexes.is_empty() {
```
rparrett, line 301: It seems like it should be possible to allow "image collection tilesets" as well as normal tilesets in a single map, provided each layer exclusively uses one or the other type of tileset.

But this code causes `TilemapTexture::Vector` to be used for every layer if any layer uses an image collection tileset.

rparrett:
A couple more notes:

- This previously (prior to this PR) worked with grouped layers, but now any layer in a group is skipped.

    This seems like sort of a pain to deal with properly (groups may be nested). This is just an example, so maybe we shouldn't worry about that.

- `cargo run --example tiled --features=atlas` fails to compile due to the use of `TilemapTexture::Vector`.

    I guess we need to gate some of this code on the atlas feature, and leave "image collection" tilesets unsupported when that feature is enabled.

dvogel:
I just pushed a commit that unifies the handling of single image and image collection tilesets, skipping the latter when the atlas feature is enabled. I also updated `map.tmx` to use these features.

Here is a screenshot with the atlas feature enabled and another with it disabled:

Screenshot from 2023-01-07 13-49-56

Screenshot from 2023-01-07 13-49-42

In the second screenshot you can see the castles because they use an image collection tileset.

Here's is the rendering of the level1.tmx file you linked to earlier:

Screenshot from 2023-01-07 13-53-32

I haven't had a chance to test a layer that uses multiple tilesets yet. I plan to do that soon but I've run out of time for the day.

examples/helpers/tiled.rs
```diff
@@ -218,32 +306,36 @@ pub fn process_loaded_maps(
                             }
                         }
 
+                        let texture = if tile_images.len() == 1 {
+                            TilemapTexture::Single(tile_images[0].clone_weak())
+                        } else {
+                            #[cfg(feature = "atlas")]
+                            panic!("Cannot use Tiled maps with a tileset that uses a collection of images due to incompatibility with the 'atlas' feature.");
```
dvogel, line 313: This should not really panic [because the tileset should be skipped](https://github.com/StarArawn/bevy_ecs_tilemap/pull/381/files#diff-45eb609402711b3323b6becf33654bed7e7e2dcf289f2d0f312a9f13565846e9R195-R198) earlier in the loop. However I wasn't sure how to satisfy the compiler without an even more extreme solution, such as generating a placeholder `TilemapTexture::Single()` here.

rparrett: I had an idea for cleaning up this and related code: https://github.com/rparrett/bevy_ecs_tilemap/commit/b9e707c360969163ad8108559ec7b926b91ed2b7

rparrett: Unfortunately, this turned out to be a bad idea, as it also can't compile with the atlas feature.

rparrett: Fixed that up in a new commit here: https://github.com/rparrett/bevy_ecs_tilemap/commit/b794f2d17fa7cb24495d3416dd995eaf1176782f which seems to be behaving.

This seems to be behaving, apart from some warnings that should probably be cleaned up when compiled with `atlas`

dvogel: That looks good to me. Works in my game. I like the idea of creating the `Single` and `Vector` structs earlier in the loading process. I had tried to do that but ran into some ownership issues and dialed it back for time.

rparrett (commented):
There are a couple warnings being generated still when building with the `atlas` feature. I couldn't figure out a way to do that any more elegantly than sprinkling more `#[cfg(not(feature = "atlas"))]` where rust is complaining.

bzm3r (commented):
@dvogel @rparrett  this is generally looking good to me. Feel free to ping me once you've settled down on solutions overall/CI is passing, and I'll merge it!

(thanks very much for your contributions!)

dvogel:
I was able to suppress all of the remaining cargo check errors by completely removing the `tile_image_offsets` member from `TiledMap` when the atlas feature is enabled.

rparrett (approved):
This seems to be working well on my end now. Thanks for all the effort to make this work with the `atlas` feature.

It's a bit unfortunate that this code that is likely to end up copy/pasted into users' projects won't work in those projects as-is. (due to feature gating on a `bevy_ecs_tilemap` feature). Users would need to manually fix up the feature gated code based on whether or not they need the `atlas` feature.

That seems preferable to other options
- not supporting image collection tilesets
- adding a completely separate `TiledMapPlugin` helper and feature gate in `examples/tiled.rs` instead
- ?

Perhaps we should at least add a comment at the top somewhere explaining the situation.

bzm3r:
>     * adding a completely separate `TiledMapPlugin` helper and feature gate in `examples/tiled.rs` instead

Hmm, I am thinking: this might be a pretty good idea. Speaking from an end-user perspective, it seems clearer? 

But we can leave this for another PR. For now, we can merge as is, with a comment explaining the situation.

bzm3r (changes requested):
This looks good to me overall. We can further refine it new PRs. I do like @rparrett 's suggestion that we should leave a comment explaining the situation, so that's the only change I am going to request!

dvogel:
I just pushed a comment with an explanation of how the code will need to be cleaned up when adopted. @rparrett please double-check my explanation as it took me a few reads through your comments to understand that you meant their code wouldn't compile once copied over.

> Hmm, I am thinking: this might be a pretty good idea. Speaking from an end-user perspective, it seems clearer?

To be honest I was a initially a little lost re: this feature. I didn't realize it was an option until well into trying to update this helper. I think even with the comment I added to the top of the helper a new user may struggle to understand what to do since the atlas feature isn't advertised up front. As best I can tell it is an optimization to load sprite sheets into the GPU more efficiently than could be done otherwise. I don't understand how it is incompatible with the alternative tilemap texture modes though. My current impression is that the feature could be removed without much downside. I assume that's not actually true, but whatever I'm missing is likely a good piece to add to the documentation :)

Long term it seems like maybe this could become full-featured enough to remain in the bevy_ecs_tilemap crate, re-exported for end-users to `use` in their projects. That would bring the compile time choices along with it.

rparrett:
> As best I can tell it is an optimization to load sprite sheets into the GPU more efficiently than could be done otherwise. I don't understand how it is incompatible with the alternative tilemap texture modes though. My current impression is that the feature could be removed without much downside. I assume that's not actually true, but whatever I'm missing is likely a good piece to add to the documentation :)

[The feature is not well documented](https://github.com/StarArawn/bevy_ecs_tilemap/issues/294). I left a comment in there that might be helpful for understanding the atlas feature. But the gist is that it's there for compatibility with webgl2.

> their code wouldn't compile once copied over.

Right, yeah. It's unfortunate that I didn't realize that sooner. I think the comment generally looks good.

Although now that it's all spelled out, it feels quite bad to ask users to do all that. Looking forward to when the `atlas` feature is eventually removed, I am liking the idea of splitting the file more and more.

We should definitely follow up on that before the next release if there's consensus about that.

bzm3r (approved):

bzm3r:
Thank you so much, dvogel and rparrett. 🙏 

It will be nice indeed when we can yeet `atlas`....

PR #385, opened 2023-01-27
Treat TilePos as a value type in APIs

johanhelsing:
Passing `TilePos` by value often makes the API a lot simpler to use (considering `TilePos` already implements `Copy`.

This also makes  `TileStorage::get` et. all generic over `Into<TilePos>`, which means you can now do:

```rust
let tile = tile_storage.get(uvec!(12, 34));
```

### DIFF
--- a/src/helpers/hex_grid/axial.rs
+++ b/src/helpers/hex_grid/axial.rs
@@ -35,8 +35,8 @@ pub struct AxialPos {
     pub r: i32,
 }
 
-impl From<&TilePos> for AxialPos {
-    fn from(tile_pos: &TilePos) -> Self {
+impl From<TilePos> for AxialPos {
+    fn from(tile_pos: TilePos) -> Self {
         AxialPos {
             q: tile_pos.x as i32,
             r: tile_pos.y as i32,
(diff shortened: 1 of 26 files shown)


src/helpers/hex_grid/axial.rs
```diff
@@ -35,8 +35,8 @@ pub struct AxialPos {
     pub r: i32,
 }
 
-impl From<&TilePos> for AxialPos {
-    fn from(tile_pos: &TilePos) -> Self {
+impl From<TilePos> for AxialPos {
```
bzm3r, line 38: I have heard that doing `From<AsRef<TilePos>` allows us to retain flexibility of calling `from` on a reference *or* value of something. Is this true? In this case, that might be a good option here?

src/tiles/storage.rs
```diff
@@ -26,15 +26,17 @@ impl TileStorage {
     /// position.
     ///
     /// Panics if the given `tile_pos` doesn't lie within the extents of the underlying tile map.
-    pub fn get(&self, tile_pos: &TilePos) -> Option<Entity> {
+    pub fn get<T: Into<TilePos>>(&self, tile_pos: T) -> Option<Entity> {
```
bzm3r, line 29: should this be part of a separate PR, where maybe this pattern is used more often?

src/helpers/hex_grid/axial.rs
johanhelsing: Yeah, I think that could be nicer. Maybe I'll play a bit more around with that and maybe split this into separate prs

johanhelsing:
Sorry, I don't I have the capacity to follow up on this one right now. Maybe I'll open new more targeted PRs later.

PR #392, opened 2023-02-19
Only call get_mut on Images that need their TextureDescriptors changed.

### LINKED ISSUE #391: Tile textures get re-copied every frame, even when not updated
I've been noticing some bad performance when using a large amount of textures (1000+) for my tiles, and looking at RenderDoc, it seems mostly related to copying texture data into images on the GPU. I'd expect that to happen once, when the tilemap is initialized, and then only have updated textures copied over after that, but it seems to happen every frame for every texture even when nothing has changed.

This also seems to happen in all of the examples I've tried. The tell-tale sign is `vkCmdCopyBufferToImage` for the tile images:

image

(Image from RenderDoc for the `tiled` example)

This also seems to happen for all the other examples I've tried (including `basic`), so I'm not sure that it relates to a particular `TilemapTexture` type

tec27:
Calling get_mut triggers an AssetEvent::Modified, and this function is called every frame, for every image that is a TileTexture. These events are used by Bevy internally to re-copy texture data to the GPU, so the effect of this was that every TileTexture would be uploaded to the GPU every frame, regardless of whether it changed.

Fixes #391.

### DIFF
--- a/src/map/mod.rs
+++ b/src/map/mod.rs
@@ -153,15 +153,19 @@ impl TilemapTexture {
     /// Sets images with the `COPY_SRC` flag.
     pub fn set_images_to_copy_src(&self, images: &mut ResMut<Assets<Image>>) {
         for handle in self.image_handles() {
-            if let Some(mut image) = images.get_mut(handle) {
+            // NOTE: We retrieve it non-mutably first to avoid triggering an `AssetEvent::Modified`
+            // if we didn't actually need to modify it
+            if let Some(image) = images.get(handle) {
                 if !image
                     .texture_descriptor
                     .usage
                     .contains(TextureUsages::COPY_SRC)
                 {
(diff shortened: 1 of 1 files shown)


tec27:
Just to also note, this function does have *some* cost to run every frame (increasing with the number of tile textures you have), and is almost always a no-op after it has run the first time. It might be nice to only do this check if the asset has changed

rparrett:
I can reproduce this issue, and have verified that this fixes it.

This needs a `cargo fmt` but otherwise looks good to me.

bzm3r:
@tec27 CI is failing on this (cargo fmt)

Also, can we rebase main (which includes bevy 0.10 changes now) onto this branch?

tec27:
Ran `cargo fmt` and rebased

rparrett (approved):
Thanks!

PR #393, opened 2023-02-20
Use clone_weak when cloning tile textures in extraction to avoid making unnecessary work for Bevy's unused asset check.

tec27:
On my usage of this with a large number of tile textures, I noticed a pretty significant amount of time being spent in Bevy's unused asset check every frame (1-2ms) even though no assets were becoming unused. It checks this by processing all the handle reference count changes during the frame, which indicated to me it must be something using `clone` on Handles.

Moving the `extract` code to use weak Handles instead of strong ones (which seems safe to me, the Tilemap should be holding a strong Handle to the texture anyway) removed that function from my trace entirely, so it's a pretty nice performance boost for larger amounts of textures.

### DIFF
--- a/src/render/extract.rs
+++ b/src/render/extract.rs
@@ -294,7 +294,7 @@ pub fn extract(
                     spacing: *data.3,
                     grid_size: *data.4,
                     map_type: *data.5,
-                    texture: data.6.clone(),
+                    texture: data.6.clone_weak(),
                     map_size: *data.7,
                     visibility: data.8.clone(),
                     frustum_culling: *data.9,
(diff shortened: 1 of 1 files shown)


rparrett (approved):
This makes sense / looks good to me and seems to follow what bevy does in its own extract systems.

bzm3r:
@tec27 Can you rebase main onto this branch, and check to see if everything still works? We can then merge this change in. Sorry for the delay.

tec27:
No worries! Rebased and everything still seems to work.

PR #396, opened 2023-03-05
Bevy 0.10

johanhelsing:
Makes basic example work with latest bevy

- [ ] switch to bevy 0.10 when out
- [ ] fix the rest of the examples
- [ ] figure out if I got the mip_levels right

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ atlas = []
 render = []
 
 [dependencies]
-bevy = { version = "0.9", default-features = false, features = [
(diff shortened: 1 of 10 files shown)


geieredgar:
I am sorry for not having chosen a better name, but I did already port the codebase to bevy 0.10 here: #390. I've added the (0.10) suffix now. All examples are already ported. I used `gpu_texture.mip_level_count()` for the mip level count instead of a hard coded `1`.

johanhelsing:
Ah, bummer. Don't know why I didn't see it. I'll let you continue on yours then 👍

Issue #353, opened 2022-11-23
Support for tilemap textures without padding


rparrett:
Is there a way to use tilemap textures like this one:

https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/assets/tiles_with_spacing.png

There there is some amount of padding between tiles, but none "outside" of the tiles?

That files comes from this repo, so presumably this was possible at one time. But it doesn't seem to be used in any of the current examples.

It would be nice if `bevy_ecs_tilemap` like bevy supported a separate value for this "offset" and "padding."

rparrett:
Ah, this behavior may be `atlas` / array texture specific at the moment, see https://github.com/StarArawn/bevy_ecs_tilemap/issues/352#issuecomment-1324419665

PR #404, opened 2023-03-22
fix tile position correcting spacing

fabiao:


fabiao:
(i as f32 % columns).floor() * (tile_size.x + spacing.x) + **spacing.x**
the last spacing sum cause left pacing be drawn (same for y)

rparrett:
Thanks for the PR. 

However, this PR would break other tilemaps, which you can see by running the `spacing` example.

In `bevy_ecs_tilemap` "spacing" means not only the distance from tile to tile, but also the distance between the tiles and the edge of the texture.

For the moment, your texture needs to include that padding.

There's a feature request here: https://github.com/StarArawn/bevy_ecs_tilemap/issues/353. Supporting a separate "offset" and "padding" is a bit more involved.

fabiao:
Yep, thanks for explanation.

Il giorno gio 30 mar 2023 alle ore 17:30 Rob Parrett <
***@***.***> ha scritto:

Issue #403, opened 2023-03-22
Hexagonal tmx map will not render correctly


darkautism:
## In Tiled Editor:
![image](https://user-images.githubusercontent.com/3898040/226789706-bdc18eeb-ee44-4553-aabc-d8ef14b08806.png)
##In Bevy ecs tilemap:
![image](https://user-images.githubusercontent.com/3898040/226789662-ad8b4193-7eca-4953-bc13-c71488f176d2.png)

## This is t1.png
![image](https://user-images.githubusercontent.com/3898040/226789888-0f7e7c4a-c90f-4f46-b1fd-15c630667f93.png)
## This is tmx file
```
<?xml version="1.0" encoding="UTF-8"?>
<map version="1.10" tiledversion="1.10.0" orientation="hexagonal" renderorder="right-down" width="18" height="10" tilewidth="48" tileheight="54" infinite="0" hexsidelength="26" staggeraxis="y" staggerindex="odd" nextlayerid="2" nextobjectid="1">
 <tileset firstgid="1" name="t1" tilewidth="48" tileheight="54" tilecount="1" columns="1">
  <image source="tiles/t1.png" width="48" height="54"/>
 </tileset>
 <layer id="1" name="Layer 1" width="18" height="10">
  <data encoding="base64" compression="zstd">
   KLUv/WDQAeUBAFAAAQEAAQEBAQEBHaCAfWADSeohACiwECWHwVncQwOrX2HuohFAI8zdiZaXlO5S+myG50vazEDnIKfTOw==
  </data>
 </layer>
</map>
```
## This is bevy_ecs_tilemap version
tiled = "0.11.0"

rparrett:
Are you able reproduce this with the in-repo `tiled` example and `tiled` 0.10?

Also, it looks like perhaps you've uploaded the first screenshot twice by mistake rather than `t1.png`.

darkautism:
Reupload t1.png
 in-repo helper has path bug if tmx more deeper
 ![image](https://user-images.githubusercontent.com/3898040/226794657-41fee2b9-39b9-43db-899b-b9668624c235.png)
 It has x not flip bugs(I already fix)
![image](https://user-images.githubusercontent.com/3898040/226795247-545e8297-8fa9-4df4-bb05-cb8e0b97e463.png)

darkautism:
Maybe i should extend TilemapBundle's map_size to double times and change every TilePos's x ?

rparrett:
[...]

PR #405, opened 2023-03-23
Add stagger map support

darkautism:
Support staggered map.
See #403

### DIFF
--- a/examples/helpers/tiled.rs
+++ b/examples/helpers/tiled.rs
@@ -12,7 +12,7 @@
 //   * When the 'atlas' feature is enabled tilesets using a collection of images will be skipped.
 //   * Only finite tile layers are loaded. Infinite tile layers and object layers will be skipped.
 
-use std::io::BufReader;
+use std::path::Path;
 
 use bevy::{
     asset::{AssetLoader, AssetPath, LoadedAsset},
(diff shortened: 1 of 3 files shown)


darkautism:
`Is the file not found error from BytesResourceReader user-facing at all?`

I think user will not fact  file not found error because I only use this loader load a tmx. If a tmx use extern png it will load behind.
But i'm not sure TSX tileset. rs-tiled has some tileset's attribute bug so it cannot read tsx correctly.

examples/helpers/tiled.rs
```diff
@@ -259,9 +312,28 @@ pub fn process_loaded_maps(
                         };
 
                         let map_type = match tiled_map.map.orientation {
-                            tiled::Orientation::Hexagonal => {
-                                TilemapType::Hexagon(HexCoordSystem::Row)
-                            }
+                            tiled::Orientation::Hexagonal => match tiled_map.map.stagger_axis {
+                                StaggerAxis::X => match tiled_map.map.stagger_index {
+                                    StaggerIndex::Even => {
+                                        // TODO: Why Even is Odd?
```
StarArawn, line 318: I believe this is because of how bevy handles Y coordinates. In tiled map editor Y down goes into the positive and in bevy Y down goes into the negative. When we load the tiles in we end up flipping the entire tilemap along the Y axis using: `mapped_y = tiled_map.map.height - 1 - y;`. Because of this we also need to "flip" the stagger.

rparrett: That makes sense. I'm not sure what to do about that other than add a comment to explain the situation.

PR #406, opened 2023-03-25
Fixed 3d iso example..

StarArawn:
Fixes the 3D iso stuff. This uses a basic but slow technique to render "3D" isometric tilemaps. This same technique should also work for "3d" hexagons as well. This works by moving each tile row into it's own "render chunk" and applying a Z value for said chunks based off of Y sort.

I think the biggest question is the API here. We might want create a new creation function on `TilemapRenderSettings` instead to create the desired layout.


examples/3d_iso.rs
```diff
@@ -0,0 +1,44 @@
...
+        .insert_resource(TilemapRenderSettings {
+            // Map size is 12x12 so we'll have render chunks that are:
+            // 12 tiles wide and 1 tile tall. 
+            render_chunk_size: UVec2::new(3, 1),
+        })
+        .add_plugins(
+            DefaultPlugins
+                .set(WindowPlugin {
+                    primary_window: Some(Window {
+                        title: String::from("Color Example"),
```
rparrett, line 28:
```suggestion
                        title: String::from("3D Isometric Example"),
```

rparrett:
Cool!

I'd be in favor of merging as-is. It doesn't look like this should affect anyone who's not trying to do 3d iso stuff, and improvements can be made later.

Couple non-blocking comments

- What's `iso_map2` doing for us?
- The visuals in the `colors` example are pretty intense now.

StarArawn:
> Cool!
> 
> I'd be in favor of merging as-is. It doesn't look like this should affect anyone who's not trying to do 3d iso stuff, and improvements can be made later.
> 
> Couple non-blocking comments
> 
> * What's `iso_map2` doing for us?
> * The visuals in the `colors` example are pretty intense now.

I'll delete `iso_map2` was just a smaller test file which helped to debug stuff better. I'll undo the colors change. I think I was trying to solve some webGL2 issues with texture arrays by doing that. 👍

rparrett (approved):
lgtm

PR #408, opened 2023-03-28
Fixed issues with z-index by hiding the y-sort behind a flag.

StarArawn:


rparrett:
I like this change, but it's unfortunate that it's a breaking change.

I personally feel like the right call might be to remove the y-sorting for a `0.10.1` release and then do this in the main branch / for an eventual `0.11`. I think that would work out better for most people. Very few should be relying on it right now, and it seems to be broken in some cases anyhow.

StarArawn:
> I like this change, but it's unfortunate that it's a breaking change.
> 
> I personally feel like the right call might be to remove the y-sorting for a `0.10.1` release and then do this in the main branch / for an eventual `0.11`. I think that would work out better for most people. Very few should be relying on it right now, and it seems to be broken in some cases anyhow.

Yeah, I was thinking maybe it was okay to sneak by, but anyone using `TilemapRenderSettings` would probably get a little upset.

src/map/mod.rs
```diff
@@ -30,6 +30,7 @@ pub struct TilemapRenderSettings {
     ///
     /// Smaller chunk sizes will benefit tilemaps which change frequently.
     pub render_chunk_size: UVec2,
+    pub y_sort: bool,
```
rparrett, line 33:
```suggestion
    /// If true, uses the chunk's `z` and `y` values when sorting during rendering.
    ///
    /// When using this option with layered tilemaps, `z` values for layers should be separated by
    /// at least `1.0` units.
    ///
    /// `render_chunk_size`'s `z` value should be `1` when using this for 3d isometric tilemaps.
    pub y_sort: bool,
```

rparrett (approved):

PR #416, opened 2023-04-11
Fix conversions between axial and offset hex coords

### LINKED ISSUE #414: Hexagon generation seems to not working correctly
I'm using `generate_hexagon` to select tiles that are in range for player to see. But this is what I get:
image

It's supposed to be just a hexagon, but every other row below the center is shifted!

Here are code snippets of the project:
```rs
fn chart_map(
    player: Query<(&MapPos, &ChartRange), With<PlayerVehicle>>,
    mut tiles: Query<(&mut TileVisibility, &TilePos, &TilemapId)>,
    chunks: Query<&Chunk>,
) {
    let (player_pos, chart_range) = player.single();
    let tiles_in_chart_range: Vec<RowEvenPos> =
        generate_hexagon(player_pos.pos.into(), chart_range.0)
            .into_iter()
            .map(Into::into)
            .collect();
    for (mut tile_vis, tile_pos, tilemap_id) in tiles.iter_mut() {
        let chunk = chunks.get(tilemap_id.0).unwrap();
        let global_tile_pos = global_from_chunk_and_local(chunk.pos, *tile_pos);
        if tiles_in_chart_range.contains(&global_tile_pos) {
            *tile_vis = TileVisibility::Visible
        } else if matches!(*tile_vis, TileVisibility::Visible) {
            *tile_vis = TileVisibility::Charted
        }
    }
}

pub fn global_from_chunk_and_local(chunk: IVec2, local: TilePos) -> RowEvenPos {
    RowEvenPos {
        q: chunk.x * TILEMAP_CHUNK_SIZE.x as i32 + local.x as i32,
        r: chunk.y * TILEMAP_CHUNK_SIZE.y as i32 + local.y as i32,
    }
}
```

And also the project is on github: https://github.com/JohnTheCoolingFan/sands_of_merkhyl/tree/b69881fde365a79716ed8cb02999443350f8a832

rparrett:
Fixes #414

This fixes conversions between `AxialPos` and (`RowEvenPos`, `RowOddPos`, `ColEvenPos`, `ColOddPos`) when coordinates are negative.

See https://www.redblobgames.com/grids/hexagons/#conversions

This seems correct, but it's currently unclear to me if similar work needs to be done in the shaders.

### DIFF
--- a/src/helpers/hex_grid/axial.rs
+++ b/src/helpers/hex_grid/axial.rs
@@ -108,8 +108,11 @@ impl From<AxialPos> for RowOddPos {
     #[inline]
     fn from(axial_pos: AxialPos) -> Self {
         let AxialPos { q, r } = axial_pos;
-        let delta = r / 2;
-        RowOddPos { q: q + delta, r }
+
+        RowOddPos {
+            q: q + (r - (r & 1)) / 2,
+            r,
+        }
     }
 }
 
@@ -117,8 +120,11 @@ impl From<RowOddPos> for AxialPos {
     #[inline]
     fn from(offset_pos: RowOddPos) -> Self {
         let RowOddPos { q, r } = offset_pos;
-        let delta = r / 2;
-        AxialPos { q: q - delta, r }
+
+        AxialPos {
+            q: q - (r - (r & 1)) / 2,
+            r,
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


rparrett:
I uncommented the stuff in the CI workflow that actually runs tests.

rparrett:
I can't remember if some of these offsets **should** differ from redblob's numbers due to bevy's coordinate system or whatever.

We should double check that before merging.

PR #425, opened 2023-05-06
Update Tiled crate and related example to 0.11.0

Bushstar:
- Updated the Tiled crate to 0.11.0.
- Creates BytesResourceReader and implemented the ResourceReader trait for the new tiled::Loader behaviour.
- Renames TileLayer to Tiles.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ ldtk_rust = { version = "0.6" }
 rand = "0.8"
 env_logger = "0.9"
 serde_json = { version = "1.0" }
-tiled = { version = "0.10.2", default-features = false }
+tiled = { version = "0.11.0", default-features = false }
 
 [dev-dependencies.bevy]
 version = "0.10"
--- a/examples/helpers/tiled.rs
+++ b/examples/helpers/tiled.rs
@@ -12,7 +12,9 @@
 //   * When the 'atlas' feature is enabled tilesets using a collection of images will be skipped.
 //   * Only finite tile layers are loaded. Infinite tile layers and object layers will be skipped.
 
-use std::io::BufReader;
+use std::io::Cursor;
+use std::path::Path;
+use std::sync::Arc;
 
 use bevy::{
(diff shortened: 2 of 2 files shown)


rparrett:
There's another (abandoned?) implementation here: #405 but I prefer yours because of the limited scope and error handling.

The other PR goes through the effort of bypassing Tiled's cache with a `NoopResourceCache` which I suspect is the right thing to do but I haven't looked closely at the issue. Tilemaps can be pretty large though and I would guess that not holding them in memory in more places than we have to would be worthwhile.

Bushstar:
Note that the the cache is not a new addition to Loader and defaulted to DefaultResourceCache before, this PR does not affect cache behaviour. You can review the changes to Tiled here where a Reader was added to Loader in the link below. I'd be inclined to leave this as is and leave users to explicitly change this behaviour if required.

https://github.com/mapeditor/rs-tiled/commit/e176b4b77f4eb3ea6041bab3630dc83b4942babf

rparrett:
Makes sense to me.

rparrett (approved):

Bushstar:
When and if this PR gets merged I'd like to create a new PR based on the changes in this one to support TMX maps with TSX tilesets. The changes are shown in the commit below.

https://github.com/Bushstar/bevy_ecs_tilemap/commit/a46094d2d679e85ccc908e2d1a38071f30ad74c9

The problem with Bevy and TSX tile maps seems to be that when rs-tiled tries to load the TMX map and comes across the TSX file it will try to read it from the bytes again and will then error as those bytes are the same TMX file just read.

In the above commit I've updated the code to check the Path, if it includes a .tsx extension then it should try and load from disk.

The "assets" directory is hard coded in this commit, we cannot get the path from LoadContext as the AssetIO member is private outside of bevy::asset. Could create a CustomAssetIO, but this then starts to get complex, especially for example code.

bjorn (approved):

StarArawn (approved):

PR #428, opened 2023-05-13
Custom shaders

StarArawn:
This PR adds custom shaders to `bevy_ecs_tilemap` in the form of a `MaterialTilemap` trait. Users can create their own materials like:

```rust
#[derive(AsBindGroup, TypeUuid, Debug, Clone, Default)]
#[uuid = "31575692-a956-4762-98e2-5d457f552d0a"]
pub struct MyMaterial {}

impl MaterialTilemap for MyMaterial {
    fn fragment_shader() -> bevy::render::render_resource::ShaderRef {
        "custom_shader.wgsl".into()
    }
}
```

This works fundamentally in the same way as bevy's `Material2d` or `Material3d` traits.

Migration Guide:
`TilemapBundle` should be replaced with `StandardTilemapBundle`.
`TilemapBundle<T: MaterialTilemap>` is now used for custom materials.


rparrett:
This is pretty neat.

Seeing a couple issues:

- Seems like the chunking example is misbehaving when moving away from the origin.
- The docs for `TilemapBundle` and `StandardTilemapBundle` are identical. Would be nice for those to explain how they differ and perhaps link to each other.

And just an opinion:

- I think `TilemapBundle` is a better name for `StandardTilemapBundle`. It seems like users not doing fancy stuff shouldn't need to think about / migrate anything.

StarArawn:
Closes: #443

PR #429, opened 2023-05-14
Add support for TSX tilesets

Bushstar:
Workaround to support Tiles TSX tilesets files. When a TSX file extension is encountered in read_from that file will be read from the path and the bytes ignored.

Bevy gives a byte stream to the TMX file, when Tiled comes across a TSX dependency from that file it will try and read it from the same stream which results in the same TMX file being read. The Tiled crate does not yet support loading child dependencies separately.

Some discussion where I initially queried this as a bug, it was mentioned that Tiled may at some point be updated with Bevy in mind, until then we have this workaround. Works locally and I can access custom tile properties from the TSX tileset.

https://github.com/mapeditor/rs-tiled/issues/265#issuecomment-1537424140

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ bevy = { version = "0.10", default-features = false, features = [
     "bevy_asset",
     "bevy_sprite",
 ] }
+
+futures-lite = "1.13.0"
 log = "0.4"
 regex = "1.5.4"
 
--- a/examples/helpers/tiled.rs
+++ b/examples/helpers/tiled.rs
@@ -12,12 +12,12 @@
 //   * When the 'atlas' feature is enabled tilesets using a collection of images will be skipped.
 //   * Only finite tile layers are loaded. Infinite tile layers and object layers will be skipped.
 
-use std::io::Cursor;
+use std::io::{Cursor, Error, ErrorKind, Read};
 use std::path::Path;
 use std::sync::Arc;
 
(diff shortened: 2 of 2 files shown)


NHodgesVFX:
One thing I noticed when trying to load a tsx was the image path that the tsx referenced was being doubled up on 

For example 

In the assets directory  the tileset image is saved as "maps/demo/TestTileTest.png"  however, when loading, it attempts to load it from /maps/demo/maps/demo/TestTileTest.png resulting in the warning

`WARN bevy_asset::asset_server: encountered an error while reading an asset: path not found: ` this causes the map to not be loaded.

The cause is on line 159 `let tile_path = tmx_dir.join(&img.source);` &img.source already contains the correct path so the join here causes the doubling up. Removing join call solves this issue.

I am quite new to bevy so not sure if removing the join call will have unintended consequences but for the one map I loaded it seems to work fine.

StarArawn:
I have no interest in merging in new features to the tiled asset loader. My reasoning is that its outside of the scope of what this project is. The tiled examples primary function is only meant to be an example on how you could load in tiled data. Which is why its located in the example folder and not part of the crate itself. I'm more than happy to work with someone on creating a `bevy_tiled` crate. :+1:

stevepryde:
fyi I started a new crate and applied this PR to it. 
https://github.com/stevepryde/bevy_ecs_tiled

PR #431, opened 2023-05-22
Track changed Image assets and use them to update image caches.

### LINKED ISSUE #430: Hot reloading tilesets does not work
I am on Windows using version 0.10.0 (I also tried HEAD of the master branch), and hot reloading the tilesets doesn't seem to work!

I looked into this myself, and it looks like this is because the image_bind_groups are not updated [here](https://github.com/StarArawn/bevy_ecs_tilemap/blob/0e71c8af09ae303916200fd5fdc1d36c9dc1a64a/src/render/queue.rs#LL164C2-L164C2).

In addition, the TextureArrayCache (when not using atlas) does not invalidate its GPU images when the asset changes.

andriyDev:
Previously, since the bind groups do not get updated every frame, modified image assets would not be rebound. This caused hot reloading to no longer work. Similarly, the TextureArrayCache (when not using atlas) would not invalidate modified assets, so the textures would become out-of-date.

Note there can be other reasons that an asset can be modified, so this should keep the assets more up-to-date in general.

Fixes #430.

### DIFF
--- a/src/render/mod.rs
+++ b/src/render/mod.rs
@@ -129,7 +129,7 @@ impl Plugin for TilemapRenderingPlugin {
                 StandardTilemapMaterial::default(),
             );
 
-        app.init_resource::<ModifiedImageHandles>()
+        app.init_resource::<ModifiedImageIds>()
             .add_systems(Update, collect_modified_image_asset_events);
     }
 
@@ -257,7 +257,7 @@ impl Plugin for TilemapRenderingPlugin {
                 (
                     extract::extract,
                     extract::extract_removal,
-                    extract_resource::<ModifiedImageHandles>,
+                    extract_resource::<ModifiedImageIds>,
                 ),
             )
             .add_systems(
@@ -274,7 +274,7 @@ impl Plugin for TilemapRenderingPlugin {
             .init_resource::<SpecializedRenderPipelines<TilemapPipeline>>()
             .init_resource::<MeshUniformResource>()
             .init_resource::<TilemapUniformResource>()
-            .init_resource::<ModifiedImageHandles>();
+            .init_resource::<ModifiedImageIds>();
 
         render_app.add_render_command::<Transparent2d, DrawTilemap>();
     }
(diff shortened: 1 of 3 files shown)


andriyDev:
Rebased onto main to fix cargo clippy which was broken at HEAD at the time.

gerasim13:
Can it be merged to main? This commit fixes the issue with updating texture on a map.

andriyDev:
I will rebase this once #440 goes through. I still need a review though for the overall structure of this solution.

andriyDev:
I've rebased this onto main after #440. I tested this with a little example and it still works.

andriyDev:
@StarArawn Is there anything I should change?

gerasim13:
@StarArawn can it be merged?

StarArawn:
@rparrett can you review this for me?

rparrett:
@andriyDev is it possible to use https://docs.rs/bevy/latest/bevy/render/render_asset/struct.ExtractedAssets.html instead of doing the manual collecting/extraction?

rparrett:
It seems like the answer to the above is probably no.

Sorry,but I don't think I can provide a meaningful review. I don't think I'll be able to get up-to-speed on this area of `bevy_ecs_tilemap` / `bevy` any time soon.

But this does fix hot reloading for me and doesn't seem to cause other issues as far as I can tell.

andriyDev:
In theory I think I could extract the `EventReader<AssetEvent<Image>>` resource, then do the collection in a render system, but since we need the modified images in `queue_material_tilemap_meshes` and `remove_modified_textures`, we'd still need the `ModifiedImageHandles` resource. It's not clear to me that that would be worth it? But if that seems nicer, I can make that change!

src/render/mod.rs
```diff
@@ -242,6 +244,9 @@ impl Plugin for TilemapRenderingPlugin {
             Shader::from_wgsl
         );
 
+        app.init_resource::<ModifiedImageHandles>()
```
rparrett, line 247: Should these main-world resources / systems be in in `build` rather than `finsih`?

andriyDev: Yep! I didn't even realize this was in `finish` haha. Done.

rparrett (approved):
Approving with the caveat: https://github.com/StarArawn/bevy_ecs_tilemap/pull/431#issuecomment-1732356953

rparrett:
Looks like CI is just failing due to new `let-else` formatting in rust 1.72 in an unrelated file.

andriyDev:
What should I do in regards to https://github.com/StarArawn/bevy_ecs_tilemap/pull/431#issuecomment-1732363616 ? I'm happy to change it if that looks cleaner. They both look roughly the same to me.

rparrett:
> They both look roughly the same to me.

I agree. Seems fine as-is IMO.

StarArawn (approved):

StarArawn:
@andriyDev Can we get the conflicts fixed up so we can get this merged in? :+1:

andriyDev:
Alright I rebased onto main. I also used AssetIds in place of Handles, since AssetIds seem to be the more "canonical" identifier for assets. I tested this locally and it still works for hot reloading!

@StarArawn I think this is good to go :)

rparrett:
Hey @andriyDev, could I bug you to update the branch again? It looks like a relatively straightforward rebase. We should be in a better position to try to get this merged now.

If not, I'd be happy to adopt the PR.

rparrett:
Merged the adopted PR. Thanks `@andriyDev`.

PR #433, opened 2023-05-29
Fixed error in chunking example

DefaultLP:
When converting a f32 to an i32 it gets rounded to the integer value that is the next closest to zero.
Meaning both [0.5, 0.5] and [-0.5, -0.5] would be rounded to [0, 0].
In the example, the [0, 0] chunk was actually four times as large and the chunks did not update when moving in range -1..1


DefaultLP:
I'm thinking of reworking the example as a whole. Like the `despawn_outofrange_chunks` uses distance and magic numbers instead of in which chunk the player is.

PR #434, opened 2023-06-08
Derive `FromReflect` for `TileTextureIndex` and more

Carnagion:
Pretty self-explanatory.

I required `TileTextureIndex` to implement `FromReflect`, as I was using it with `bevy_proto`, which requires it for turning components into `Schematic`s.

While I was at it, I also derived `FromReflect` for other types that implemented `Component` and `Reflect`, since it makes sense for most components on entities to be reflectable.


StarArawn (approved):
This looks good assuming CI passes. 👍

PR #440, opened 2023-07-09
Bevy 0.11 Update

### LINKED ISSUE #439: Bevy 0.11 Support
I'm starting work on Bevy 0.11 support unless somebody else already has a good start.  I'm just going to follow the change log strictly.

mwbryant:
All examples are tested and work on my machine TM.  

Some concerns I have are with the shader imports, they work but I'm not great with wgsl so I had to duplicate the locations instead of using an import due to the changes in Bevy 0.11.  Also constants were giving an exotic error message so I moved them into the functions that used them.

Also anisotropy_clamp must be set now, I don't personally know the details but it was previously none and wgpu has the default as 1 so I went with that. 

Change detection magic timer matches the recommended time in the migration guide.

Otherwise nothing major, a couple of undocumented fixes I had to find the PRs for and silent failures but nothing too bad.

Closes #439


src/render/shaders/tilemap_vertex.wgsl
```diff
@@ -1,44 +1,49 @@
-#import bevy_ecs_tilemap::common
+#import bevy_ecs_tilemap::common VertexInput, tilemap_data
```
rparrett, line 1: We need `tilemap_data` for the `atlas` feature.
```
#import bevy_ecs_tilemap::common tilemap_data
```

rparrett (changes requested):
Thanks for picking this up. I have a couple notes based on a quick skim and an attempt to use the crate with wasm.

I think we should add the feature `webgl2` to the first `dev-dependencies` section for wasm support in our examples.

examples/3d_iso.rs
```diff
@@ -1,4 +1,6 @@
-use bevy::prelude::*;
+use std::time::Duration;
```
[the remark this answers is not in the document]
mwbryant: Done!

src/render/shaders/tilemap_vertex.wgsl
mwbryant: I'm unclear what you are asking to be changed here?  Does this import not already include tilemap_data or are you saying it should not?

rparrett:
With regard to the duplicated stuff after the shader import changes, I think something like this works.

https://gist.github.com/rparrett/b122349a0ed2a13dd3d303cd085c50db

src/render/shaders/tilemap_vertex.wgsl
rparrett: My bad, I meant to place that comment in `tilemap_fragment.wgsl`.

src/render/shaders/tilemap_fragment.wgsl
```diff
@@ -1,8 +1,11 @@
-#import bevy_ecs_tilemap::common
```
rparrett, line 1: We need `tilemap_data` too for the `atlas` feature.

mwbryant: Ah good catch, done!

mwbryant:
That does work and makes the final result better than the original code

rparrett:
Could you check out the shader errors in `column_even_hex.wgsl` and `column_odd_hex.wgsl`?

I'm seeing them in the `hexagon_column` and `hex_neighbors` examples. (Press space to toggle modes, some are broken)

mwbryant:
Those examples should work now and I did a spot check over all the shaders and they all have any imports I can see.  Is there a good way to force a verification or something?  These runtime shader issues are fragile.

Also I keep moving const declaration into the functions that use them because of this shader error:
``` Composable module identifiers must not require substitution according to naga writeback rules: `bevy_ecs_tilemap::column_even_hex::SQRT_3` ```

I have no idea what this could mean or what the right solution to that is but moving the constants solves the error.

**Review by rparrett — APPROVED** [review]
> I have no idea what this could mean or what the right solution to that is but moving the constants solves the error.

I'm also a bit confused about this and looking at `naga_oil`'s source didn't really clear that up for me.

The workaround seems fine IMHO.

File: examples/accessing_tiles.rs [hunk 4]
```diff
@@ -160,9 +160,9 @@ fn main() {
...
                 .set(ImagePlugin::default_nearest()),
         )
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(update_map)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, update_map)
```
BernardIgiri, line 166: These calls can be combined. The `add_systems` method allows you to add multiple systems in one call.

examples/animation.rs
```diff
@@ -139,12 +139,12 @@ fn main() {
...
-        .add_plugin(LogDiagnosticsPlugin::default())
-        .add_plugin(FrameTimeDiagnosticsPlugin::default())
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_startup_system(create_background)
-        .add_startup_system(create_animated_flowers)
-        .add_system(helpers::camera::movement)
+        .add_plugins(LogDiagnosticsPlugin::default())
+        .add_plugins(FrameTimeDiagnosticsPlugin::default())
+        .add_plugins(TilemapPlugin)
```
BernardIgiri, line 144: These calls to `add_plugins` can be combined as the new method allows you to pass in multiple plugins at once.

examples/animation.rs
```diff
@@ -139,12 +139,12 @@ fn main() {
...
+        .add_systems(Startup, startup)
+        .add_systems(Startup, create_background)
+        .add_systems(Startup, create_animated_flowers)
```
BernardIgiri, line 147: These `add_systems` calls can be combined.

examples/basic.rs
```diff
@@ -109,9 +109,9 @@ fn main() {
...
             ..default()
         }).set(ImagePlugin::default_nearest()))
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(swap_texture_or_hide)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, swap_texture_or_hide)
```
BernardIgiri, line 115: These `add_systems` calls can  be combined.

examples/bench.rs
```diff
@@ -56,10 +56,10 @@ fn main() {
...
             ..Default::default()
         })
-        .add_plugin(LogDiagnosticsPlugin::default())
-        .add_plugin(FrameTimeDiagnosticsPlugin::default())
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
+        .add_plugins(LogDiagnosticsPlugin::default())
+        .add_plugins(FrameTimeDiagnosticsPlugin::default())
+        .add_plugins(TilemapPlugin)
```
BernardIgiri, line 61: These `add_plugins` calls can  be combined.

examples/chunking.rs
```diff
@@ -122,11 +122,11 @@ fn main() {
...
+        .add_plugins(TilemapPlugin)
         .insert_resource(ChunkManager::default())
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(spawn_chunks_around_camera)
-        .add_system(despawn_outofrange_chunks)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, spawn_chunks_around_camera)
+        .add_systems(Update, despawn_outofrange_chunks)
```
BernardIgiri, line 130: These `add_systems` calls can  be combined.

examples/game_of_life.rs
```diff
@@ -106,9 +106,9 @@ fn main() {
...
                 .set(ImagePlugin::default_nearest()),
         )
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(update)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, update)
```
BernardIgiri, line 112: These `add_systems` calls can  be combined.

examples/hexagon_column.rs
```diff
@@ -128,9 +128,9 @@ fn main() {
...
                 .set(ImagePlugin::default_nearest()),
         )
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(swap_mesh_type)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, swap_mesh_type)
```
BernardIgiri, line 134: These `add_systems` calls can  be combined.

examples/hexagon_row.rs
```diff
@@ -127,9 +127,9 @@ fn main() {
...
                 .set(ImagePlugin::default_nearest()),
         )
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(swap_mesh_type)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, swap_mesh_type)
```
BernardIgiri, line 133: These `add_systems` calls can  be combined.

examples/move_tile.rs
```diff
@@ -67,9 +67,9 @@ fn main() {
...
                 .set(ImagePlugin::default_nearest()),
         )
-        .add_plugin(TilemapPlugin)
-        .add_startup_system(startup)
-        .add_system(helpers::camera::movement)
-        .add_system(swap_pos)
+        .add_plugins(TilemapPlugin)
+        .add_systems(Startup, startup)
+        .add_systems(Update, helpers::camera::movement)
+        .add_systems(Update, swap_pos)
```
BernardIgiri, line 73: These `add_systems` calls can  be combined.

examples/hex_neighbors_radius_chunks.rs
```diff
@@ -541,18 +541,18 @@ fn main() {
...
-        .add_system(hover_highlight_tile_label.after(swap_map_type))
-        .add_system(update_radius.after(hover_highlight_tile_label))
-        .add_system(highlight_neighbor_labels.after(update_radius))
+        .add_systems(Startup, (spawn_chunks, apply_deferred).chain().in_set(SpawnChunksSet))
+        .add_systems(Startup, spawn_tile_labels.after(SpawnChunksSet))
+        .add_systems(First, (camera_movement, update_cursor_pos).chain())
+        .add_systems(Update, swap_map_type)
+        .add_systems(Update, hover_highlight_tile_label.after(swap_map_type))
+        .add_systems(Update, update_radius.after(hover_highlight_tile_label))
+        .add_systems(Update, highlight_neighbor_labels.after(update_radius))
```
BernardIgiri, line 556: I haven't tested this, but I believe that these `add_systems` calls can be combined as well. You should be able to combine the Startup systems in one call, and the Update systems in another call I believe.

BernardIgiri (commented):
I see some opportunity to combine some of these calls to `add_systems` and `add_plugins`. Might as well take advantage of the new API functionality now.

StarArawn (approved):

examples/hex_neighbors_radius_chunks.rs
StarArawn: A great idea! I'm not going to hold the PR up for it though. :) Feel free to open another PR and I'll merge it for sure. 👍

PR #442, opened 2023-07-20
Allow examples to be built for WebGPU and add instructions

rparrett:
I sort of forgot that WebGPU is a thing because browser support is so limited. But folks probably want to use it, so we shouldn't include `bevy/webgl2` by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,8 +46,7 @@ features = [
     "bevy_winit",
     "bevy_text",
     "bevy_sprite",
-    "file_watcher",
-    "webgl2",
+    #"file_watcher",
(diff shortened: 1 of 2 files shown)


rparrett:
#443 443

StarArawn:
@rparrett can you fix the conflicts? :)

StarArawn (approved):

rparrett:
In #497 we apparently un-commented the `file_watcher` feature in `Cargo.toml`, so wasm wasn't building at all.

I just re-commented it, but we could also fix that by adding a wasm target specific feature list.

StarArawn (approved):

PR #451, opened 2023-07-25
Minimal fix for chunks with entity generation > 0 not rendering

### LINKED ISSUE #450: Render chunks with tilemap Entity generation > 0 are not visible
```
AdapterInfo { name: "Apple M1 Max", vendor: 0, device: 0, device_type: IntegratedGpu, driver: "", driver_info: "", backend: Metal }
SystemInfo { os: "MacOS 13.4.1 ", kernel: "22.5.0", cpu: "Apple M1 Max", core_count: "10", memory: "64.0 GiB" }
```

bisected to 7f5113adf8851d654adc145e54b062bcb15d12b5

I remember noticing this here: https://github.com/StarArawn/bevy_ecs_tilemap/pull/428#issuecomment-1548712692 but can't recall if there was a fix that maybe didn't survive a merge or something.

## Repro

`cargo run --example chunking`

Hold the`w` key. Newly spawned chunks after the initial setup mostly do not display.

chunking

rparrett:
This is a minimal changeset that fixes #450.

This probably shouldn't be merged as-is, because it makes a bit of a mess with some duplicated data. I don't fully understand this code and don't feel equipped to fix it properly at the moment.

### DIFF
--- a/src/render/chunk.rs
+++ b/src/render/chunk.rs
@@ -38,6 +38,7 @@ impl RenderChunk2dStorage {
         &mut self,
         tile_entity: Entity,
         tile_pos: UVec2,
+        chunk_entity: Entity,
         position: &UVec4,
         chunk_size: UVec2,
         mesh_type: TilemapType,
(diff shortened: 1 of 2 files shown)


StarArawn (approved):
This looks good. 👍

rparrett:
Okay, yeah. I am thinking that this fix is solid. I may follow up later because it seems like

- `RenderChunk2d::id` is unused
- And perhaps some cleanup could be done in `RenderChunk2dStorage`.

Issue #456, opened 2023-08-13
Choice of hex tile size in examples prevents pixel-perfect tiling


ulfhermann:
The hex tiles are 15x17 pixels. Since when hex-tiling the rows or columns are offset by half a tile and neither the width nor the height are integer-divisible by 2, you necessarily get a pattern where every other "column" or "row" looks somewhat off. It took me a while to figure out that this was not some rendering deficiency but simply a matter of choosing even numbers for the sizes.

ulfhermann:
An important information to be considered in this context is that bevy automatically calculates a scale factor for each window, which may be something like 1.08333 if you have exotically sized screens. You can override that thing:

```
app.add_plugins(
            DefaultPlugins
                .set(WindowPlugin {
                    primary_window: Some(Window {
                        title: String::from("Hexagon Column Example"),
                        resolution: WindowResolution::default().with_scale_factor_override(1.0),
                        ..Default::default()
                    }),
                    ..default()
                })...
```

Furthermore, you probably want to turn off anti-aliasing: 

`app.insert_resource(Msaa::Off);`

I'm preparing a patch that gets rid of the sqrt(3) calculations and rather calculates the tile offsets from the given grid size using divisions by 2 and 4. That way a nicer grid size (like 28x32, for example) actually has an effect. If you do the sqrt(3) calculations, the calculation errors add up and your pixels will still be "off" in some places.

PR #463, opened 2023-08-26
Use grid size based arithmetics for offsets in hex tile layout

ulfhermann:
By refraining from using sqrt(3) based arithmetics we get the tiles aligned to the provided textures and grid size. This is not mathematically exact, but arguably what the user wants when providing tiles of specific size.

The central realization here is that neighboring tiles on the "diagonal" axis are always offset by a half tile size in one direction and a 3/4 tile size in the other. So, by providing tiles with sizes divisible by 4, you can get pixel perfect alignment of tiles.

The calculations are also quite a bit simpler this way.

See #456 for some further comments on this.


StarArawn (commented):
This math has changed quite a bit from when I first wrote it. Arguably for the better. I can't say I know enough about the math specifically or have the time to dive in to see if this is correct in all user cases. If someone else is willing to review this PR we can merge it assuming it looks good to go.

ulfhermann:
I've updated the patch and resolved the conflict.

src/render/shaders/column_even_hex.wgsl
```diff
@@ -5,13 +5,7 @@
 
 // Gets the screen space coordinates of the bottom left of an isometric tile position.
```
teohhanhui, line 6: Not related to this PR, but all the comments seem to be copy and paste... Haha.

src/render/shaders/row_odd_hex.wgsl
```diff
@@ -5,13 +5,7 @@
...
 // Gets the screen space coordinates of the bottom left of an isometric tile position.
 fn hex_row_tile_pos_to_world_pos(pos: vec2<f32>, grid_width: f32, grid_height: f32) -> vec2<f32> {
-    let SQRT_3: f32 = 1.7320508;
-    let HALF_SQRT_3: f32 = 0.8660254;
-    let ROW_BASIS_X: vec2<f32> = vec2<f32>(1.0, 0.0);
-    let ROW_BASIS_Y: vec2<f32> = vec2<f32>(0.5, HALF_SQRT_3);
-
-    let unscaled_pos = pos.x * ROW_BASIS_X + pos.y * ROW_BASIS_Y;
-    return vec2<f32>(grid_width * unscaled_pos.x, ROW_BASIS_Y.y * grid_height * unscaled_pos.y);
+    return vec2<f32>(grid_width * (pos.x + pos.y / 2.0), grid_height * pos.y * 0.75);
```
teohhanhui, line 8: I think naming these in variables as before would help, e.g.

```suggestion
    let ROW_BASIS_X: f32 = 0.5;
    let ROW_BASIS_Y: f32 = 0.75;

    return vec2<f32>(grid_width * (pos.x + pos.y * ROW_BASIS_X), grid_height * pos.y * ROW_BASIS_Y);
```

PR #468, opened 2023-09-12
Implement `Add<Vec2>` for Vec2-like structs

Piturnah:
There are many structs that have a very Vec2-like structure to them and I have found myself wanting the convenience of using `+` to just add a `Vec2` to them. This PR implements this operation for those types. Let me know if I missed some!

### DIFF
--- a/src/map/mod.rs
+++ b/src/map/mod.rs
@@ -6,6 +6,22 @@ use bevy::{
     prelude::{Component, Entity, Handle, Image, Reflect},
 };
 
+/// For the convenience in implementing `Add<Vec2>` for Vec2-like structures which are common in
+/// this module.
+macro_rules! impl_add_vec2 {
+    ($ty:ty $(, $inner_ty:ty)?) => {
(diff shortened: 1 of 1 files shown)


src/map/mod.rs
```diff
@@ -6,6 +6,22 @@ use bevy::{
...
 
+/// For the convenience in implementing `Add<Vec2>` for Vec2-like structures which are common in
+/// this module.
+macro_rules! impl_add_vec2 {
+    ($ty:ty $(, $inner_ty:ty)?) => {
+        impl std::ops::Add<Vec2> for $ty {
+            type Output = Self;
+            fn add(self, rhs: Vec2) -> Self::Output {
+                Self {
+                    x: self.x + rhs.x $(as $inner_ty)?,
```
StarArawn, line 17: I'm not sure I like the `as u32` here since it hides potential unwanted side effects in this macro that would be difficult for a user to debug.

PR #474, opened 2023-10-08
Fix animated tiles not being updated when AnimatedTile changes

rparrett:
Fixes an issue discussed in #473.

Also adds a pause button to the example.

PR #480, opened 2023-10-15
v0.11.1 release

StarArawn:


rparrett:
CI fix is here: #475 

#474 and #478 are very small fixes that you may want to throw in.

#431 is a fix I tested but don't understand well and is maybe not very high impact.

StarArawn:
> CI fix is here: #475
> 
> #474 and #478 are very small fixes that you may want to throw in.
> 
> #431 is a fix I tested but don't understand well and is maybe not very high impact.

Already fixed CI issues, I don't have much more time and I'd like to get the release out. We can always role a v0.11.2

rparrett:
Sounds good.

PR #481, opened 2023-10-15
Fix tile color srgbness

### LINKED ISSUE #479: Tile Color is wrong, because of sRGB/linear not being handled correctly
When using the TileColor to colorize the sprites, the colors look wrong. They look pale/light, which is indicative of missing gamma correction. sRGB is probably not set correctly in the tilemap shader.

Here is a screenshot rendered using Bevy Sprites (correct colors):

Screenshot 2023-10-14 at 17 34 00

Here is a screenshot rendered using Tilemap (incorrect colors):

Screenshot 2023-10-14 at 17 30 31

I can apply the following workaround when setting the TileColor value, to manually "correct" the RGB values:

```rust
        let my_color = /* ... */;
        let rgba = my_color.as_linear_rgba_f32();
        tile_color.0 = Color::rgba(rgba[0], rgba[1], rgba[2], rgba[3]);
```

Now the color looks correct, the same as with Bevy Sprites. This indeed confirms that the issue is the tilemap rendering is missing sRGB handling.

rparrett:
Fixes #479

Not as familiar as I would like to with the rendering code. Is it okay to change the type of the color field in `PackedTileData`? I could leave it the same, but the rest of the code would likely be uglier.

### DIFF
--- a/src/render/chunk.rs
+++ b/src/render/chunk.rs
@@ -167,7 +167,7 @@ pub struct PackedTileData {
     pub visible: bool,
     pub position: Vec4,
     pub texture: Vec4,
-    pub color: Vec4,
+    pub color: [f32; 4],
 }
 
 #[derive(Clone, Debug)]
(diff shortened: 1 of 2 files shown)


StarArawn:
Functionally I Vec4 would be the same.

StarArawn (approved):

PR #489, opened 2023-11-13
Support Bevy v0.12.0

divark:
## Summary
This PR makes the necessary changes in order for this library to be bevy v0.12.0 compliant. This closes #488.

## Testing
1. Clone the `bevy_ecs_tilemap` repository as-is.
2. Run all of the examples as described in the `README.md`.
3. Clone my version of `bevy_ecs_tilemap`
4. Run `cargo check` to verify there are no compilation issues.
5. Run all of the examples as described in the `README.md`.
6. Observe there are (hopefully) no differences between the original.

PR #497, opened 2023-12-17
Upgade to bevy 0.12 

StarArawn:
This is just a small bug fix onto of the great work that @divark has done so far.

Thank you for all of the hardwork on this one! 

Closes: #489


TMeerhof:
Nice job! Thank you
Was a release made? crates.io still shows 0.11.
https://crates.io/crates/bevy_ecs_tilemap

PR #501, opened 2023-12-24
Fix tiled helper map location propagation

### LINKED ISSUE #500: TilePos world location is not aligned with the drawn location in the tiled helper
e.g. the following would return the world location ignoring the transform from here:
https://github.com/StarArawn/bevy_ecs_tilemap/blob/b08a5d997867d5e5a760e296dc38e435c6b268cc/examples/helpers/tiled.rs#L378-L383
or in the `TiledMapBundle` and defaulting to the bottom left at 0,0.
```rust
tile_pos.center_in_world(grid_size, map_type)
```
resulting in the world position of the tile entity not aligning with the drawn location

cbrown1234:
fixes #500

### DIFF
--- a/examples/helpers/tiled.rs
+++ b/examples/helpers/tiled.rs
@@ -18,6 +18,7 @@ use std::sync::Arc;
 
 use bevy::{
     asset::{io::Reader, AssetLoader, AssetPath, AsyncReadExt},
+    hierarchy::BuildChildren,
     log,
     prelude::{
         Added, Asset, AssetApp, AssetEvent, AssetId, Assets, Bundle, Commands, Component,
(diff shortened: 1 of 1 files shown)


examples/helpers/tiled.rs
```diff
@@ -365,6 +369,9 @@ pub fn process_loaded_maps(
                                     })
                                     .id();
                                 tile_storage.set(&tile_pos, tile_entity);
+                                // Make tiles transform relative to layer transform, 
+                                // to ensure they have correct world_pos aligned with drawn location
+                                commands.entity(layer_entity).add_child(tile_entity);
```
StarArawn, line 374: Tiles don't have bevy transforms. This wont do what you want it to here. Due to how bevy's transform propagation works it would be incredible slow to have each tile have a bevy transform so I can't recommend that.

PR #505, opened 2024-01-20
Add const to some functions.

poweron0102:
It just make some `::new()` function to be const.

allowing it to be call on const definitions like this:

```rust
// Old
const SIZE: TilemapSize = TilemapSize{ x: 256, y: 256 };

// New
const SIZE: TilemapSize = TilemapSize::new(256, 256);
```  
And some compile time optimizations.


StarArawn (approved):

StarArawn:
I fixed the clippy issue elsewhere. I'll merge this. :+1:

PR #508, opened 2024-01-31
Upgrade to Bevy 0.13

rparrett:
This was an easy one and largely mechanical. But I am not super familiar with `EntityMapper`, so it would be great if someone could help test that part.


StarArawn (approved):

linonetwo:
Can we have a release? Thanks.

PR #512, opened 2024-02-10
Moved render settings onto entity.

StarArawn:
## Description
This allows a user to customize render settings per tilemap entity.

## Why is this useful?
It allows users to specify render chunk sizes per tilemap/layer. If a user has a tilemap that doesn't change often they can gain performance by setting the chunk size to be very large reducing draw calls and they can mix another tilemap layer that might update more frequently on top of the more static map.

### DIFF
--- a/examples/bench.rs
+++ b/examples/bench.rs
@@ -43,10 +43,6 @@ fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
     });
 }
 
-fn display_t_entities(query: Query<Entity, With<Transform>>) {
-    dbg!(query.iter().count());
-}
-
 fn main() {
     App::new()
         .add_plugins(
@@ -64,6 +60,6 @@ fn main() {
         .add_plugins(FrameTimeDiagnosticsPlugin)
         .add_plugins(TilemapPlugin)
         .add_systems(Startup, startup)
-        .add_systems(Update, (helpers::camera::movement, display_t_entities))
+        .add_systems(Update, helpers::camera::movement)
         .run();
 }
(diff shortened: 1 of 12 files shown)


StarArawn:
Realized I didn't update the examples, working on that now..

examples/bench.rs
```diff
@@ -39,6 +39,10 @@ fn startup(mut commands: Commands, asset_server: Res<AssetServer>) {
     });
 }
 
+fn display_t_entities(query: Query<Entity, With<Transform>>) {
```
rparrett, line 42: Was this change included intentionally?

rparrett:
Any reason not to set a reasonable default render chunk size and include this in the bundles?

examples/bench.rs
StarArawn: nope. :+1:

examples/chunking.rs
```diff
@@ -117,11 +121,7 @@ fn main() {
                 })
                 .set(ImagePlugin::default_nearest()),
         )
-        // `TilemapRenderSettings` be added before the `TilemapPlugin`.
-        .insert_resource(TilemapRenderSettings {
-            render_chunk_size: RENDER_CHUNK_SIZE,
-            ..Default::default()
-        })
+        // `TilemapRenderS
```
rparrett, line 124: Typo

rparrett (approved):
Looks good to me, and I tested all the examples.

It occurred to me that I have no idea how to test when the `render` feature is not enabled though.

Noticed one more little typo.

StarArawn:
> Looks good to me, and I tested all the examples.
> 
> It occurred to me that I have no idea how to test when the `render` feature is not enabled though.
> 
> Noticed one more little typo.

You can test that by using the `--no-default-features` cargo parameter.

rparrett:
> You can test that by using the --no-default-features cargo parameter.

Oh, yeah, I just meant that we don't have any examples that utilize this "mode." They all require the `render` feature, and I just sort of don't know what is even expected to work.

PR #518, opened 2024-03-02
Added more serde support

StarArawn:


src/tiles/mod.rs
```diff
@@ -53,6 +53,15 @@ impl From<UVec2> for TilePos {
     }
 }
 
+impl From<IVec2> for TilePos {
```
rparrett, line 56: Non-blocking commentary: feels a bit footgunny

rparrett (approved):
I find Bevy's serde / reflect stuff quite confusing and don't really have any way of testing but I don't see anything wrong.

src/tiles/mod.rs
StarArawn: Ah yeah I was testing some stuff I'll remove.

Issue #521, opened 2024-03-13
Bevy 0.13 race condition resulting in tiles not rendering


MeoMix:
Hello! I'm trying to look into this but I don't have a ton of ideas just yet. I wanted to report it just to promote awareness.

After updating to Bevy 0.13 and pointing to main on bevy_ecs_tilemap I intermittently have an issue where no tiles render. 

Here are some screenshots:

Incorrect:
![image](https://github.com/StarArawn/bevy_ecs_tilemap/assets/1380995/76ca2f3e-bc87-4e52-838f-8b392670e137)

Correct:
![image](https://github.com/StarArawn/bevy_ecs_tilemap/assets/1380995/af186240-9713-4979-a2b4-6c77b362ffce)

Other assets which don't rely on bevy_ecs_tilemap render fine. The issue reproduces on both native builds and WASM builds. 

Additionally, I can confirm that the handle is strong and is fully loaded by the time I give it to `TilemapBundle` as a texture.

I reviewed the PR which brought bevy_ecs_tilemap up to 0.13 and didn't immediately spot anything concerning. I am going to try and test the examples locally now and see if I can reproduce the issue with any of them.

I wonder if this has to do with changes to `apply_deferred` such that it attempts to apply itself automatically? Although I would think that would cause less issues to occur, but reduce performance, rather than introducing new race conditions. Since it reproduces in WASM I don't think it has anything to do with multithreading.

MeoMix:
https://github.com/MeoMix/bevy_bug_example/blob/master/src/main.rs

This minimally reproduces the issue. It's some conflict with the bevy_egui render feature and bevy_ecs_tilemap but I haven't tracked it down further yet.

MeoMix:
Okay, I put up a fix!

For historical tracking purposes (I reuse my bug example repo sometimes), here's relevant code to reproduce the issue:

```rust
// This code should draw a white square in the center of the window if it runs properly.
// One in every 5-10 runs no white square will render.
// If the `multithreaded` feature is enabled in bevy's Cargo.toml then the square "always" renders (if it's a timing issue it becomes sufficiently difficult to surface with multithreading)
use bevy::prelude::*;
use bevy::render::{Render, RenderApp, RenderSet};
use bevy_ecs_tilemap::prelude::*;

fn main() {
    App::new()
    .add_plugins((DefaultPlugins, TilemapPlugin))
    .add_systems(Startup, (spawn_camera, spawn_tilemap).chain())
[...]

PR #523, opened 2024-03-19
Fix ambiguous `queue_material_tilemap_meshes` ordering

MeoMix:
It was possible for `queue_material_tilemap_meshes` to run prior to `prepare` which would result in nothing rendering.

This would only occur in scenarios where other code inserted a system dependent on `mut commands: Commands` into the `Queue` set. The `bevy-egui` library was doing so and was thus introducing behavior where `bevy-ecs-tilemap` would intermittently not render anything due to improper system ordering.

Fixes: https://github.com/StarArawn/bevy_ecs_tilemap/issues/521

### Additional Details

https://github.com/djeedai/bevy_hanabi/pull/297 Bevy Hanabi experienced a similar issue. Their fix looks similar to this proposed fix.

The Bevy notes hint that this sort of change is necessary, too:

image

This is **not** directly an issue with automatically applying command buffers. It is related, but the key issue here is one of improper system ordering. The existing codebase _happened_ to work but was not _guaranteed_ to work. It worked only because the scheduler was choosing to resolve ambiguous system ordering in a way which was compatible with `bevy_ecs_tilemap` architecture. This invited the problem where third party code could introduce systems that would affect the scheduler's ordering decision and result in rendering failures.

### DIFF
--- a/src/render/material.rs
+++ b/src/render/material.rs
@@ -30,6 +30,7 @@ use super::{
     chunk::{ChunkId, RenderChunk2dStorage},
     draw::DrawTilemapMaterial,
     pipeline::{TilemapPipeline, TilemapPipelineKey},
+    prepare,
     queue::{ImageBindGroups, TilemapViewBindGroup},
 };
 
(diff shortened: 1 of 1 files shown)


StarArawn (approved):

StarArawn:
@MeoMix can you run cargo fmt on this and then it's good to merge!

MeoMix:
@StarArawn Feel free to merge? I don't have the permission to do so

Issue #305, opened 2022-10-06
Add `TilePos::from_cursor_pos`


alice-i-cecile:
As demonstrated in the `neighbors.rs` example, you need to do some fancy math to get the world position from the cursor position.

We should have an API which handles this for you, to supplement `TilePos::from_world_pos`.

StarArawn:
I agree that it's complicated but I don't think it is something that `bevy_ecs_tilemap` should provide. It feels out of scope. I think this is something that would be great to have native bevy support for. Alternatively we could possibly look into using `bevy_mod_picking` by adding 2d support.

frederickjjoubert:
> I agree that it's complicated but I don't think it is something that `bevy_ecs_tilemap` should provide. It feels out of scope. I think this is something that would be great to have native bevy support for. Alternatively we could possibly look into using `bevy_mod_picking` by adding 2d support.

I disagree that it is out of scope. I think it is a common enough use case with games that use Tiles that `bevy_ecs_tilemap` should include it.

StarArawn:
> > I agree that it's complicated but I don't think it is something that `bevy_ecs_tilemap` should provide. It feels out of scope. I think this is something that would be great to have native bevy support for. Alternatively we could possibly look into using `bevy_mod_picking` by adding 2d support.
> 
> I disagree that it is out of scope. I think it is a common enough use case with games that use Tiles that `bevy_ecs_tilemap` should include it.

It is a common use case, but camera's are controlled by bevy not by `bevy_ecs_tilemap`. I think an example of how it works is sufficient enough until bevy can provide a generic way of handling this. I think my biggest point here is that the work should be done in bevy and then used in the `bevy_ecs_tilemap` examples. If I offer a very specific mouse to tile function that does little for any of bevy's 2D stuff and will only work for `bevy_ecs_tilemap`, but if the work was done in bevy directly then it can be reused to also allow sprite mouse picking within bevy.

rparrett:
I think this was largely solved when Bevy added `viewport_to_world_2d`.

It's super simple to get this now.

PR #529, opened 2024-04-19
Added an example to iso_diamond example for handling tile clicks in respect to the 2d camera

ThunderSizzle:
I added an example to iso_diamond's example for working with the cursor position with the tile map. This permits the user to click on any tile to change the color while the example is running. Left clicking increments the tile's color index, and right clicking decrements the tile's color index.

This is to offer an example for this issue: [https://github.com/StarArawn/bevy_ecs_tilemap/issues/305Issue 305](https://github.com/StarArawn/bevy_ecs_tilemap/issues/305) 

--
Part of the reason for this example is that it took me the better part of a week to figure out how to do this. I wanted an example of this to be accessible.

### DIFF
--- a/examples/iso_diamond.rs
+++ b/examples/iso_diamond.rs
@@ -1,4 +1,5 @@
 use bevy::prelude::*;
+use bevy::window::PrimaryWindow;
 use bevy_ecs_tilemap::prelude::*;
 mod helpers;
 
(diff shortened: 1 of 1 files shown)


rparrett:
This seems to be implementing Bevy's [`viewport_to_world_2d`](https://docs.rs/bevy/latest/bevy/prelude/struct.Camera.html#method.viewport_to_world_2d) manually. I think #305 probably predates that addition to Bevy, and can be closed now.

We also have a [mouse_to_tile](https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/examples/mouse_to_tile.rs) example that works with iso diamond.

ThunderSizzle:
@rparrett: Thanks for reviewing. I see now my misunderstanding of the mouse_to_tile example. I didn't realize that "CursorPos" was self-contained in the example. I attempted using TilePos::from_world_pos using the cursor position from the window, which obviously didn't work, hence how I guess ended up rebuilding "viewport_to_world_2d" that I didn't know existed. 

That simplifies my usage quite a bit. Thank you!

PR #536, opened 2024-06-05
Use license instead of license-file

### LINKED ISSUE #490: License shown as "non-standard" on crates.io
On [crates.io](https://crates.io/crates/bevy_ecs_tilemap), your license is shown as "non-standard" even though you're using the MIT license. I'm not entirely sure why, but my guess is it's because you've set `license-file` in your `Cargo.toml`. I don't think that's necessary if you've already specified `license`.

jgayfer:
The license-file is intended to be used for nonstandard licenses, per the cargo docs.

> If a package is using a nonstandard license, then the license-file field
> may be specified in lieu of the license field.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields

Closes #490


rparrett (approved):
Makes sense, thanks.

PR #540, opened 2024-07-13
Move src/map/mod.rs to src/map.rs

### LINKED ISSUE #520: Instead ov `src/map/mod.rs`, it should just be `src/map.rs`.
Unless there are plans to add more to the `src/map` directory, I believe `src/map/mod.rs` should become `src/map.rs`.  
This change would, structurally, make more sense, as there is not really a reason for `mod.rs` to be all alone in its own directory.

If there is any reason why it is like this, let me know;  I am interested to hear why it is designed like this.

stevepryde:
This simplifies the module structure a tiny bit.

Fixes #520


ChristopherBiscardi (approved):
seems straightforward and previously desired as indicated by the issue linked

rparrett (approved):

CalinZBaenen:
Let's go!  🎉

PR #547, opened 2024-08-01
adopted: Track changed Image assets and use them to update image caches

rparrett:
Original PR text:

> Previously, since the bind groups do not get updated every frame, modified image assets would not be rebound. This caused hot reloading to no longer work. Similarly, the TextureArrayCache (when not using atlas) would not invalidate modified assets, so the textures would become out-of-date.
> 
> Note there can be other reasons that an asset can be modified, so this should keep the assets more up-to-date in general.

Fixes https://github.com/StarArawn/bevy_ecs_tilemap/issues/430
Alternative to #431 

## Testing

I modified `tiles.png` and observed hot reloading occurring with:

`cargo run --example basic --features=bevy/file_watcher`
`cargo run --example basic --features=atlas --features=bevy/file_watcher`


ChristopherBiscardi (approved):
Looks good to me. I spot-checked the basic example with and without atlas.

https://github.com/user-attachments/assets/b68c5c3d-fbf3-4afb-9814-de4fd67f3896

src/render/texture_array_cache.rs
```diff
@@ -342,3 +344,24 @@ impl TextureArrayCache {
...
+
+    texture_cache
+        .textures
+        .retain(|texture, _| texture_is_unmodified(texture));
+    texture_cache
+        .meta_data
+        .retain(|texture, _| texture_is_unmodified(texture));
+    texture_cache.prepare_queue.retain(texture_is_unmodified);
+    texture_cache.queue_queue.retain(texture_is_unmodified);
+    texture_cache.bad_flag_queue.retain(texture_is_unmodified);
```
ChristopherBiscardi, line 366: only strange thing I noticed was we don't seem to actually be using `bad_flag_queue` anywhere. What is this for?

rparrett: fascinating. I don't think it has ever been used. It seems to have originated here: https://github.com/StarArawn/bevy_ecs_tilemap/commit/020d10ea3c54dfacd406c873940b8bfb3cf37654

assets/tiles.png
ChristopherBiscardi, line 1: the image is also modified, but we can fix that whenever.

rparrett: ack, let me get that now

ChristopherBiscardi (approved):

src/tiles/storage.rs
```diff
@@ -47,7 +47,7 @@ impl TileStorage {
     /// Gets a tile entity for the given tile position, if:
     /// 1) the tile position lies within the underlying tile map's extents *and*
     /// 2) there is an entity associated with that tile position;
-    /// otherwise it returns `None`.
+    ///    otherwise it returns `None`.
```
ChristopherBiscardi, line 50: I think this is supposed to be a second paragraph, not a continuation of the second item

rparrett: Indeed, good catch.

PR #549, opened 2024-08-13
Update wasm instructions

rparrett:
- The rustflags are no longer necessary
- `bevy/webgl2` is already specified in `dev-dependencies`
- both examples now consistently don't use `--release`

Tested both commands to ensure they're working / produce `webgl2` and `webgpu` builds.


ChristopherBiscardi (approved):
lgtm, and works on my machine

PR #556, opened 2024-09-03
Enable `wayland` feature for `bevy` in dev

teohhanhui:
Rationale:
* https://github.com/bevyengine/bevy/issues/13753#issuecomment-2167562087
* https://github.com/bevyengine/bevy/issues/8788#issuecomment-1742192359 (it still will not work on Asahi Linux for now due to https://github.com/bevyengine/bevy/issues/13923)
* https://github.com/bevyengine/bevy/pull/10792


teohhanhui:
Uhh... Need to add the deps to CI...

teohhanhui:
Oops... Missed adding the step for the `clippy` job...

ChristopherBiscardi:
https://github.com/bevyengine/bevy/pull/10792#issuecomment-2239837897 states that

> Defaulting to Wayland will break Bevy for Wayland users with no access to Vulkan drivers.

I largely don't have an opinion on doing this for dev. It looks like the current state of things is we either potentially break one group of users or another group of users (until bevy merges the fix upstream), but wayland is the future so I'm a soft +1.

Would like to get thoughts from @rparrett though.

ChristopherBiscardi (approved):

rparrett (approved):
I'm fine with doing this before Bevy if it helps out the people who want to work on `bevy_ecs_tilemap` now.

teohhanhui:
> Defaulting to Wayland will break Bevy for Wayland users with no access to Vulkan drivers.

Actually no, it's already broken. See https://github.com/bevyengine/bevy/issues/13923

PR #560, opened 2024-09-15
Run First systems after TimeSystem in order to ensure minimum impact on delta time

VergilUa:
Systems that are added to the First schedule run before TimeSystem runs, potentially delaying proper delta time computation / frame start. 

Better solution would be to add a proper set for these systems and sort it against TimeSystem set.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -15,7 +15,7 @@
 //! - Texture array support.
 
 use bevy::prelude::{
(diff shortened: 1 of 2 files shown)


ChristopherBiscardi:
Thanks (and thanks for bringing this up in Discord)! I'm going to close this in favor of the new set in #562 , as you've mentioned.

PR #562, opened 2024-09-15
Run systems in First after Time set

ChristopherBiscardi:
If we accept that [(#4669) Res<Time> is jittery](https://github.com/bevyengine/bevy/issues/4669) then

1. Time is updated in the [TimeSystem set](https://docs.rs/bevy/0.14.2/bevy/time/struct.TimeSystem.html)
2. No work should happen before the Time is updated in a frame

This doesn't fix the upstream jitter issue, but we can reduce bevy_ecs_tilemap's potential impact on the issue by making sure any work we do in `First` is done *after* time is updated.

## Solution

1. Create a new `TilemapFirstSet` SystemSet
2. Add bevy_ecs_tilemap `First` systems to `TilemapFirstSet`
3. Order `TilemapFirstSet` systems after the `TimeSystem`.

## bevy_mod_debug_dump

system scheduling in First before change:

screenshot-2024-09-15-at-10 15 37@2x

system scheduling after change:

screenshot-2024-09-15-at-10 18 23@2x

## Migration

This shouldn't require end-user migration, but if you want to run system in First after bevy_ecs_tilemap's work, then the new SystemSet should be used.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,9 +14,13 @@
 //! - Built in animation support  – see [`animation` example](https://github.com/StarArawn/bevy_ecs_tilemap/blob/main/examples/animation.rs).
 //! - Texture array support.
 
-use bevy::prelude::{
-    Bundle, Changed, Component, Deref, First, GlobalTransform, InheritedVisibility, Plugin, Query,
-    Reflect, ReflectComponent, Transform, ViewVisibility, Visibility,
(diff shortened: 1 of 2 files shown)


rparrett (approved):
Seems fine. Also seems like it would probably be fine to dump these in `PreUpdate`.

PR #564, opened 2024-09-21
Derive `Ord` for `TilePos`

teohhanhui:
`TilePos` has a total order. See https://doc.rust-lang.org/stable/core/cmp/trait.Ord.html#corollaries

This allows using `TilePos` as key in `BTreeMap` or as value in `BTreeSet`, for example.

### DIFF
--- a/src/tiles/mod.rs
+++ b/src/tiles/mod.rs
@@ -10,7 +10,7 @@ use crate::map::TilemapId;
 use crate::TilemapSize;
 
 /// A tile position in the tilemap grid.
-#[derive(Component, Reflect, Default, Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd)]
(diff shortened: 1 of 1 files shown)


rparrett (approved):

ChristopherBiscardi:
glam's UVec2 doesn't implement Ord (or PartialOrd) and it feels weird to have an implementation here. That said, we already support PartialOrd so I'm fine adding this in but I think we should actually remove the PartialOrd in a future release

Specifically you say that "TilePos has a total order" but in the context of x/y positions I'm not sure that's accurate since there are multiple possible implementations (does x matter more or does y matter more?)

teohhanhui:
> Specifically you say that "TilePos has a total order" but in the context of x/y positions I'm not sure that's accurate since there are multiple possible implementations (does x matter more or does y matter more?)

I was just operating based on the existing derived `PartialOrd` impl, but also it'd be hardly surprising to anyone that `(x,y)` is sorted by well, `x,y`.

PR #568, opened 2024-10-25
0.15 (sort of)

bas-ie:
So this is what happens when you blithely decide that, since you're gonna need `bevy_ecs_tilemap` again soon, you may as well try to upgrade it for 0.15. Uh, yeah. Some of the rendering stuff is well over my head! :sweat_smile: 

This is _not_ a working update, but I'm leaving it here in case you can cherry pick some of the simpler commits to save you some time. I'll keep trying in the meantime, I'm learning quite a bit from it... but please don't feel the need to spend any significant time reviewing it.

It does include fixes for the following:
- https://github.com/bevyengine/bevy/issues/15716 `Handle<M>` wrapper
- https://github.com/bevyengine/bevy/pull/14257 `GpuMesh` -> `RenderMesh` _but_ the implementation is nowhere near done, there's quite a lot to unpack there that I don't grok yet
- https://github.com/bevyengine/bevy/pull/15962 `elapsed_seconds` -> `elapsed_secs`
- https://github.com/bevyengine/bevy/pull/15652 `get_or_spawn` deprecation
- https://github.com/bevyengine/bevy/pull/14909 `as_bind_group` args, needed charlotte's help with that one!
- a required components migration. I'm by no means completely convinced by the naming scheme, and would welcome alternatives. But in the end it seemed to make sense to:
  - rename `MaterialTilemap` to `TilemapMaterial`...
  - ...in order to create `MaterialTilemap` to replace `MaterialTilemapBundle`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.15.0-rc.1", default-features = false, features = [
+bevy = { version = "0.15.0-rc.2", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
@@ -34,7 +34,7 @@ tiled = { version = "0.11.0", default-features = false }
 thiserror = { version = "1.0" }
 
 [dev-dependencies.bevy]
-version = "0.15.0-rc.1"
+version = "0.15.0-rc.2"
 default-features = false
 features = [
     "bevy_core_pipeline",
@@ -51,7 +51,7 @@ features = [
 ]
 
 [target.'cfg(unix)'.dev-dependencies.bevy]
-version = "0.15.0-rc.1"
+version = "0.15.0-rc.2"
 default-features = false
 features = [
     "bevy_core_pipeline",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.14.0", default-features = false, features = [
+bevy = { version = "0.15.0-rc.1", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.15.0-rc.2", default-features = false, features = [
+bevy = { version = "0.15.0-rc.3", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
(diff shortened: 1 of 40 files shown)


ChristopherBiscardi:
thanks! your effort is much appreciated!

richchurcher:
I am being very slow, but hopefully I can get it to the point where someone who knows about rendering can take over!

richchurcher:
OK. I think that's all the low-hanging fruit. Now, how the hell do I get all this rendering working :thinking:

richchurcher:
@ChristopherBiscardi can I ask: does the rendering code need a bit of a re-think? Given the retained render world changes (`Entity`/`MainEntity`), the `GpuMesh`/`RenderMesh` changes (`MeshAllocator` etc).

The deeper I dig, the more I wonder if it needs an experienced hand to say, "yep, we still need this bit", or "no, all this can be simplified". I can keep plugging away at it, just wanted to check in first to be sure I'm actually spending time on the right things. For example, as @tychedelia points out,

> `TileMapId` wraps an `entity` that is actually `MainEntity` in the render world

and I haven't really got the first idea how to solve that :thinking: General guidance and suggestions very welcome.

richchurcher:
In the meantime, I'm going to start by working my way through the _Extract_ part of the pipeline, since that seems likely to be affected by retained rendering world.

ChristopherBiscardi:
There will be some changes for the retained rendering world but I'd like to avoid large sweeping changes otherwise, at least for the migration to 0.15-compatible. We can revisit the implementation in a larger way after getting a 0.15-compatible version out. For example, things like taking advantage of Required Components would be a breaking change that I'd like to leave for *after* the 0.15-compatible release if possible to make it easier for users of the crate to upgrade Bevy.

Re: the `.insert_or_spawn_batch` you've linked an issue that wants to deprecate it but not an actual deprecation so while we can think about removing the usage its not something that should be the top priority until it actually is deprecated or removed (it doesn't look slated for 0.15 and it also is labelled contentious).

I'll have more time to spend on this later this week. I likely won't get to much this weekend.

richchurcher:
Hrm, I could also take a peek at the minimum possible changeset, sans required components... but I think there are still gonna be rendering issues. I'll investigate.

evilenzo:
Sorry, I'm not good at rendering at all but I also want to help you a little. Just made a little pr to @richchurcher fixing remaining issues in examples: https://github.com/richchurcher/bevy_ecs_tilemap/pull/2

richchurcher:
Yeah, I've definitely arrived at my limit with the rendering stuff, it needs a more experienced hand! Once I see a few examples, I think I'll be able to approach it with more confidence.

Please feel free to close this or cherry-pick whatever you find useful.

rparrett:
Here's one possibility for migrating https://github.com/bevyengine/bevy/pull/14257: https://github.com/rparrett/bevy_ecs_tilemap/commit/3cc2cad106109ccec400be760d9e9dd8d531c2d0

@ChristopherBiscardi does that seem like a reasonable strategy?

rparrett:
The `basic` example now renders in my branch above, but it's messy and some of the migrations around render/main world entity refs seem kinda dubious.

adrien-bon:
Hi! Thanks for (mostly) fixing rendering!
I openned a PR on your branch: https://github.com/rparrett/bevy_ecs_tilemap/pull/5
It should fix render entities removal + tilemap change detection.

Issue #572, opened 2024-11-29
Add native picking support for tiles


zacryol:
With the release of `bevy 0.15`, picking has been added to the engine: https://bevyengine.org/news/bevy-0-15/#entity-picking-selection

`bevy_ecs_tilemap` should implement a backend for this functionality, allowing the selection of individual tiles.

I've previously written a crate that adds integration via `bevy_mod_picking` (https://github.com/zacryol/bevy_picking_tilemap, works with up to bevy 0.14 currently) that can be used as a starting point.

dpogorzelski:
A 5 minute hack seems to make it work again with 0.15 but i'm probably missing stuff, mostly no idea of what i'm doing :)

TheNeikos:
Here is a version for bevy 0.19, for those that are looking for something like this:

```rust
use bevy::camera::RenderTarget;
use bevy::picking::PickingSystems;
use bevy::picking::backend::HitData;
use bevy::picking::backend::PointerHits;
use bevy::picking::pointer::PointerId;
use bevy::picking::pointer::PointerLocation;
use bevy::prelude::*;
use bevy::window::PrimaryWindow;
use bevy_ecs_tilemap::prelude::*;

pub struct TilemapPickingPlugin;

impl Plugin for TilemapPickingPlugin {
    fn build(&self, app: &mut App) {
        app.add_systems(PreUpdate, tile_picking.in_set(PickingSystems::Backend));
    }
}

fn tile_picking(
    pointers_q: Query<(&PointerId, &PointerLocation)>,
    cameras_q: Query<(
        Entity,
        &Camera,
        &RenderTarget,
        &GlobalTransform,
        &Projection,
    )>,
    primary_window_q: Query<Entity, With<PrimaryWindow>>,
    tilemap_q: Query<(
        &TilemapSize,
        &TilemapGridSize,
        &TilemapTileSize,
        &TilemapAnchor,
        &TilemapType,
        &TileStorage,
        &GlobalTransform,
        &ViewVisibility,
    )>,
    mut output: MessageWriter<PointerHits>,
) {
    for (p_id, p_loc) in pointers_q
        .iter()
        .filter_map(|(p_id, p_loc)| p_loc.location().map(|l| (p_id, l)))
    {
        let Some((cam_entity, camera, _cam_target, cam_transform, _cam_ortho)) = cameras_q
            .iter()
            .filter(|(_, camera, ..)| camera.is_active)
[...]

PR #574, opened 2024-12-03
Bevy 0.15

rparrett:
This seems to be in a pretty good place now. Some things left to do:

- [X] Make CI green if it isn't
- [X] Test examples on native (macos)
- [X] Test examples with the `atlas` feature
- [X] Test examples with `webgl2` and `webgpu`
- [X] Make sure that "map removal" is working properly
- [X] Would be nice to test for any major perf regression
- [X] Would be nice to double check that we aren't leaking entities anywhere, especially when adding/removing tiles/maps.

Many thanks to @bas-ie, @evilenzo, @adrien-bon, @tychedelia.

Let me know if I missed anything important.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.15.0-rc.3", default-features = false, features = [
+bevy = { version = "0.15", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
@@ -34,7 +34,7 @@ tiled = { version = "0.11.0", default-features = false }
 thiserror = { version = "1.0" }
 
 [dev-dependencies.bevy]
-version = "0.15.0-rc.3"
+version = "0.15"
 default-features = false
 features = [
     "bevy_core_pipeline",
@@ -43,6 +43,7 @@ features = [
     "png",
     "ktx2",
     "bevy_winit",
+    "bevy_window",
     "bevy_text",
     "bevy_sprite",
     #"file_watcher",
@@ -51,7 +52,7 @@ features = [
 ]
 
 [target.'cfg(unix)'.dev-dependencies.bevy]
-version = "0.15.0-rc.3"
+version = "0.15"
 default-features = false
 features = [
     "bevy_core_pipeline",
@@ -60,6 +61,7 @@ features = [
     "png",
     "ktx2",
     "bevy_winit",
+    "bevy_window",
     "wayland",
     "x11",
     "bevy_text",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,8 @@ bevy = { version = "0.15", default-features = false, features = [
     "bevy_render",
     "bevy_asset",
     "bevy_sprite",
+    # See Bevy#16563
+    "png",
 ] }
 log = "0.4"
 serde = { version = "1", features = ["derive"], optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.14.0", default-features = false, features = [
+bevy = { git = "https://github.com/bevyengine/bevy", rev = "af865e76a323fc87dd16157e7f4d7593d02c2b95", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
@@ -34,7 +34,8 @@ tiled = { version = "0.11.0", default-features = false }
 thiserror = { version = "1.0" }
 
 [dev-dependencies.bevy]
-version = "0.14.0"
+git = "https://github.com/bevyengine/bevy"
+rev = "af865e76a323fc87dd16157e7f4d7593d02c2b95"
 default-features = false
 features = [
     "bevy_core_pipeline",
@@ -51,7 +52,8 @@ features = [
 ]
 
 [target.'cfg(unix)'.dev-dependencies.bevy]
-version = "0.14.0"
+git = "https://github.com/bevyengine/bevy"
+rev = "af865e76a323fc87dd16157e7f4d7593d02c2b95"
 default-features = false
 features = [
     "bevy_core_pipeline",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { git = "https://github.com/bevyengine/bevy", rev = "3e86787e93abb73958fd8884b77fe435e0f13333", default-features = false, features = [
+bevy = { git = "https://github.com/bevyengine/bevy", rev = "6e2f96f2224032e04e77b993ad4c06d76d49039f", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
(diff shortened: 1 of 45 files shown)


rparrett:
Looking closer at the dependency updates from https://github.com/rparrett/bevy_ecs_tilemap/pull/5, it looks like:
- bevy/wgpu are still bringing in `thiserror` `1`, so I think we should stay there anyhow.
- `env_logger` seems unused in `bevy_ecs_tilemap`. Maybe it can just be removed.
- I'd prefer to hold off on `tiled` for a followup PR.

bas-ie:
Amazing, nice work. I'll close the other branch, thanks for letting me experiment/learn!

bas-ie:
> * examples with `webgl2` and `webgpu`

I'll try to take on this bit, should be able to within the next day or so.

rparrett:
Looks like there's a major performance regression. It seems that something is going wrong with change detection and chunks are getting rebuilt every frame.

rparrett:
Okay, fixed the performance regression.

adrien-bon:
Testing with `bevy_ecs_tiled` examples, it seems we still have an issue with map removal.
From time to time, when spawning a new map in place a previous one, I see the following logs:

```
2024-12-04T18:57:56.175238Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::material::MaterialTilemapHandle<bevy_ecs_tilemap::render::material::StandardTilemapMaterial> into the following invalid entities: [5v2084#8950711844869]
2024-12-04T18:57:56.199393Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::extract::ExtractedTilemapTextureBundle into the following invalid entities: [5v2084#8950711844869]
2024-12-04T18:57:56.199471Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::material::MaterialTilemapHandle<bevy_ecs_tilemap::render::material::StandardTilemapMaterial> into the following invalid entities: [5v2084#8950711844869]
2024-12-04T18:57:56.224944Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::extract::ExtractedTilemapTextureBundle into the following invalid entities: [5v2084#8950711844869]
2024-12-04T18:57:56.225049Z ERROR bevy_ecs::system::commands: Failed to 'insert or spawn' bundle of type bevy_ecs_tilemap::render::material::MaterialTilemapHandle<bevy_ecs_tilemap::render::material::StandardTilemapMaterial> into the following invalid entities: [5v2084#8950711844869]
```

Will investigate

adrien-bon:
I can reproduce the issue if I `despawn_recursive()` a tilemap then spawn a new one. I get the `Failed to 'insert or spawn' bundle` error and my new tilemap won't render. Issue is not reproduced if I just re-insert a tilemap on the same entity.

I don't fully understand what is happening but it does not happen if I remove the `TemporaryRenderEntity` component from the `Extract*` bundles.

I dig a bit into the code, especially around extraction logic (see `src/render/extract.rs:373`).
From what I see, the `RenderEntity` I get for the tilemap is not valid anymore.

What is strange is that it works for the first frame but it does not after.
I observe the following sequence:
- spawn a tilemap, with a new `RenderEntity` = X
- `insert_or_spawn_batch(X)`
- `insert_or_spawn_batch(X)`
- ...
- despawn the tilemap, `RenderEntity` X is properly despawned
- spawn a tilemap, with a new `RenderEntity` = Y
- `insert_or_spawn_batch(Y)`
- `insert_or_spawn_batch(Y)` => error is triggered

Not sure why a `RenderEntity` I get from an `Extract<Query<T>>` would suddenly become invalid...

adrien-bon:
Got it to work by removing the `TemporaryRenderEntity` component from `ExtractedTilemapBundle` bundle.

I don't understand because this component should still be present on the tilemap entity since it is also inserted via the `ExtractedTilemapTexture` bundle on the same entity (which also have this `TemporaryRenderEntity` component).

Hopefully performances are still OK ?

PR = https://github.com/rparrett/bevy_ecs_tilemap/pull/6

rparrett:
Any chance you could provide a runnable example that reproduces this?

My thinking at the moment is that we
- need to be storing main world entity refs in `Extracted*` / eventually in `RenderChunk2dStorage`
- `insert_or_spawn` is pointless if `Extracted*` remain `TemporaryRenderEntity`, so we can remove some main/render entity confusion by just using `spawn_batch`.

(or, rework things so that we do persistent extraction, and have some separate mechanism for detecting changed tiles/maps/textures or whatever. Would probably prefer fixing up the temporary entities for now.)

(grain of salt, I don't know the renderer super well)

I had the same thought though to do what you just did as a quick workaround. I'll take a closer look in a bit.

rparrett:
Okay, I think I was misunderstanding things pretty badly -- `insert_or_spawn_batch` is not pointless, because `SyncToRenderWorld` creates the entities in the render world.

It *is* confusing though, and should probably be swapped out for `insert_batch`.

The thing that's melting my brain at the moment though is reconciling this "custom change-based extraction" with `SyncToRenderWorld` extracting every entity. It might not be a problem but it just feels weird.

Understanding this code more and more though. Will look again a little later.

rparrett:
I reverted the `TemporaryRenderEntity` stuff in favor of a new `ChangedInMainWorld` component. The queries in `prepare` skip anything with that component and that component only is cleared every frame.

Perf seems even better (~7.5ms), and map reloading seems to work again.

dpogorzelski:
Hey, in 0.14 i used to have a custom bevy_mod_picking backend to be able to react when clicking on a tile.
Considering native support for entity picking via observe in 0.15, should it be possible to .observe a TileBundle and get a click trigger to work?
Seems that a simple `.observe(on_click_print_hello);` and then:
```rust
fn on_click_print_hello(click: Trigger<Pointer<Click>>) {
    println!("{} was clicked!", click.entity());
}
```
doesn't work.

rparrett:
That isn't expected to work out of the box. Individual tiles don't have meshes or bounding boxes. 

(IMO) A picking backend would be great to add now that picking is built into Bevy. Please see #572 for related discussion.

rparrett:
Added a `despawn_tilemap` example to help with testing.

rparrett:
Okay, no idea why adding that example is breaking the headless CI, but I'll just remove it for now.

rparrett:
I spot-checked `basic`, `custom_shader`, and `tiled` with

--target=wasm32-unknown-unknown
--target=wasm32-unknown-unknown --features=atlas
--target=wasm32-unknown-unknown --features=bevy/webgpu
--target=wasm32-unknown-unknown --features=bevy/webgpu,atlas

(mac m1/chrome)

ChristopherBiscardi:
I see the review request and will start on that later today.

neocturne:
Based on this PR, the update of bevy_ecs_ldtk is also coming along nicely: https://github.com/Trouv/bevy_ecs_ldtk/pull/340

I've tested the bevy_ecs_tilemap + bevy_ecs_ldtk updates on Linux X11 and Wayland and didn't see any issues.

ChristopherBiscardi:
Reveiwed and have been using this PR for about a week now to good effect and haven't noticed any issues. Intending to approve, merge, and release later today.

ChristopherBiscardi (approved):

Issue #337, opened 2022-11-07
TilemapBundle::grid_size should scale images


Laocoon7:
I believe it should be in TilemapBundle, but I may be wrong.

Currently, `TilemapBundle::grid_size` cuts images off if smaller than `TilemapBundle::tile_size` or adds padding to the left and above if larger. The expected result is a scaled tile image which matches `TilemapBundle::grid_size`.

This would allow a custom orthographic projection mapping where world space coordinates are the same as tile position without needing to re-adjust the orthographic projection's scale per tileset size. This would also allow use of multiple tileset's where the aspect ratio is identical even if the tile sizes are not. Traditionally Bevy's Sprite::custom_size could be used to achieve the same result.

If I have misinterpreted grid_size, or a breaking change like this is not allowed, an additional field for `TilemapBundle::tile_custom_size` or the like would be requested.

bzm3r:
@Laocoon7 

In the case when grid size > tile size:

* no padding is added, it appears as if there is padding because the tile size is not scaled

In the case when grid size < tile size:

* no cutting off is done: it appears as if mages are being cut off because the tile size is not scaled

Grid size is the "logical size" of the grid positions of the underlying tile map. Put differently, it is the distance between grid positions. For example, if grid size is `(10, 20)`, the x-distance between position `(0, 0)` and `(1, 0)` will be `10` while the distance between `(0, 0)` and `(0, 1)` will be `20`.

Tile size is the "physical size" of the tiles being placed at each grid position. In particular, it is used to figure out the region of input texture that should be read from in order to visualize the tile. 

So, if scaling is required by the user, the best thing to do would be to provide a `tile_custom_size` which by default will be `tile_size`. This is something we can do.

rparrett:
Adding another component to control "tile world size" is probably the better solution, but

- Is it possible to accomplish this sort of thing by scaling the `Tilemap`?
- What about adding a `TileScale` field that could default to `Vec2::splat(1.)`

I have a concern that the TilemapBundle is becoming quite unwieldy with a bunch of components that are not super clearly distinguished that a typical user will be setting the same value.

PR #580, opened 2024-12-12
Add `in_world_tile_size` field to determine how large each tile is rendered in-world 

SirHall:
Recreated as no-one seemed to be looking at the previous PR, and the reviewer seemed to no longer be active.

As `bevy_ecs_tilemap` stands now, each tile's size in-world is tied to the pixel-size of the tile in the source atlas.  
If I however want my tiles to be of unit size or any particular scale that I require, this is not currently possible.  
With this pull request, a user can now set `physical_tile_size` to determine how large a tile should be rendered in-world.

This should also close [#337](https://github.com/StarArawn/bevy_ecs_tilemap/issues/337).


miketwenty1:
@SirHall I noticed this PR has gotten no attention publicly. Is there some official process that this PR is missing?

SirHall:
> @SirHall I noticed this PR has gotten no attention publicly. Is there some official process that this PR is missing?

I believe this ticket needs a reviewer assigned.

PR #583, opened 2024-12-17
Remove redundant time property

### LINKED ISSUE #575: `SecondsSinceStartup` resource is redundant
Noticed in #574

Need to investigate if this resource is either used for something useful, or useful to users. I suspect neither.

edit: used for animation, but this seems redundant because Bevy provides time to shaders in a global.

rparrett:
# Objective

Fixes #575

I think that this code predates time generally being available to bevy shaders in `globals`.

We can accomplish the same thing using Bevy's built-in time. This was only being used in `bevy_ecs_tilemap` for animation.

I am having trouble imagining a legit scenario where users could have been using `SecondsSinceStartup` and they couldn't accomplish whatever they were doing through less janky means. So I don't think this is likely to be breaking anyone's stuff.

The padding is no longer necessary as the resulting struct is now 16 byte aligned.

# Testing

Tested the `animation` example on
- m1 mac (native)
- chrome / webgl2


ChristopherBiscardi (approved):
makes sense to me

PR #586, opened 2024-12-23
Add `TileStorage::drain` and return removed entities in `remove`

### LINKED ISSUE #579: Consider adding `drain` method to `TileStorage`
I'd like to be able to easily despawn all the tiles in a tilemap without putting them in a Bevy hierarchy.

Also, it would be nice if `remove` returned the removed entity, if there was one.

rparrett:
Fixes #579

I was tempted to call this `remove_all` instead to remove a potential source of confusion about what is happening to the underlying storage. Still on the fence about this, so let me know what you think.

### DIFF
--- a/src/tiles/storage.rs
+++ b/src/tiles/storage.rs
@@ -86,20 +86,41 @@ impl TileStorage {
         self.tiles.iter_mut()
     }
 
-    /// Remove entity at the given tile position, if there was one, leaving `None` in its place.
+    /// Removes any stored `Entity` at the given tile position, leaving `None` in its place and
+    /// returning the `Entity`.
     ///
     /// Panics if the given `tile_pos` doesn't lie within the extents of the underlying tile map.
-    pub fn remove(&mut self, tile_pos: &TilePos) {
-        self.tiles[tile_pos.to_index(&self.size)].take();
+    pub fn remove(&mut self, tile_pos: &TilePos) -> Option<Entity> {
+        self.tiles[tile_pos.to_index(&self.size)].take()
     }
 
-    /// Remove any stored entity at the given tile position, if the given `tile_pos` does lie within
-    /// the extents of the underlying map.
+    /// Remove any stored `Entity` at the given tile position, leaving `None` in its place and
+    /// returning the `Entity`.
     ///
-    /// Otherwise, nothing is done.
-    pub fn checked_remove(&mut self, tile_pos: &TilePos) {
+    /// Checks that the given `tile_pos` lies within the extents of the underlying map.
+    pub fn checked_remove(&mut self, tile_pos: &TilePos) -> Option<Entity> {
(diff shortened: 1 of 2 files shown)


src/tiles/storage.rs
```diff
@@ -86,20 +86,41 @@ impl TileStorage {
...
-    /// Otherwise, nothing is done.
-    pub fn checked_remove(&mut self, tile_pos: &TilePos) {
+    /// Checks that the given `tile_pos` lies within the extents of the underlying map.
+    pub fn checked_remove(&mut self, tile_pos: &TilePos) -> Option<Entity> {
         if tile_pos.within_map_bounds(&self.size) {
-            self.tiles[tile_pos.to_index(&self.size)].take();
+            return self.tiles[tile_pos.to_index(&self.size)].take();
         }
+
+        None
```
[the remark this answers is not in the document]
ChristopherBiscardi: looks fine to me. the index access probably implies that this is guaranteed to be filled with None but just reading the code as an outside observer it could easily look like it was a mistake as well.

I think if we wanted to be exact we'd return `Result<Option<Entity>, BoundsError>` but checking if a tile is "in-bounds" seems trivial for users to do outside of this function if they want that information.

src/tiles/storage.rs
```diff
@@ -86,20 +86,38 @@ impl TileStorage {
...
     }
 
-    /// Remove entity at the given tile position, if there was one, leaving `None` in its place.
+    /// Removes any stored `Entity` at the given tile position, leaving `None` in its place and
+    /// returning the `Entity`.
     ///
     /// Panics if the given `tile_pos` doesn't lie within the extents of the underlying tile map.
-    pub fn remove(&mut self, tile_pos: &TilePos) {
-        self.tiles[tile_pos.to_index(&self.size)].take();
+    pub fn remove(&mut self, tile_pos: &TilePos) -> Option<Entity> {
```
ChristopherBiscardi, line 93: should we get rid of this entirely and make `checked_remove` the new `remove` while we're adding a return value? 0.16 is going to come with [fallible systems](https://github.com/bevyengine/bevy/pull/16589) too, and I can't actually think of a use case for panicking here. Its not chainable, so .unwrap() won't add length to chains of calls if panicing is preferred.

[upstream preference](https://github.com/bevyengine/bevy/issues/15979#issuecomment-2552997940) seems to be for returned errors, especially in cases of lesser-used apis.

rparrett: That makes sense.

The whole set of methods could really use a closer look though. Maybe I should remove the `remove` changes and we can discuss in a separate issue?

ChristopherBiscardi: I don't think we need to revert, but happy to discuss more generally in a separate issue. Even if we do that sort of thing the function signatures here will match and the only difference will be "no panic".

ChristopherBiscardi (approved):

PR #590, opened 2024-12-24
Fix mildly wrong doc comment

rparrett:


ChristopherBiscardi (approved):

rparrett:
Maintainers have voted that it's okay to self-approve trivial PRs like this.

PR #596, opened 2025-02-03
Make tilemap respond to `Anchor` component.

shanecelis:
Hi, thank you for all your work on this crate! It's been incredibly useful. 

I had a specific need for controlling how the tilemap was positioned. I specifically wanted it to be positioned from the top-left, but I went ahead and implemented all the positions available on [`bevy::sprite::Anchor`](https://docs.rs/bevy/latest/bevy/sprite/enum.Anchor.html). So now one can choose the anchor much the same way one chooses with a sprite. 

I added a example called "anchor" that demonstrates this feature. The tilemap is placed at the origin and only its anchor changes. Here is a movie of it.

movie of anchor positions

This is not a breaking change. Tilemaps that have no anchor component will be positioned the same as before.

I did not update the other examples that rely on `get_tilemap_center_transform()` but they can all be altered to use `Anchor::Center`. One nice property of this anchor is that the Transform does not contain any "center" information, so if you place the tilemap centered at (0,0,0), the transform will be at (0,0,0) unlike the helper function.

If this PR is accepted, and you'd like the other examples updated, let me know.

Kind regards,

Shane


ChristopherBiscardi:
ci isn't passing at the moment and this will need some docs to explain what its supposed to be doing. That said I think I like this change.

Question about how you anticipate `Anchor` being used here. It seems like `Anchor::None` places the `Transform` on the center of the first bottom-left tile, and `Anchor::BottomLeft` places it at the edge of the first bottom-left tile (the other anchors all seem to be at the edge). Do you envision the examples and suggested user usage to change to use `Anchor::Center`?

I have to check some more things, like how it behaves with multiple chunks, what other math it affects, and whether we should be using `Anchor` for this (or a custom `TilemapAnchor`). (bevy's `Text2d` and `Sprite` use `Anchor`, but I don't think `Mesh2d` does in Bevy).

shanecelis:
Hi Chris, thanks for looking at these changes!

> It seems like Anchor::None places the Transform on the center of the first bottom-left tile, and Anchor::BottomLeft places it at the edge of the first bottom-left tile

Yes, that was done to preserve the original behavior where no `Anchor` is present. Placing the anchor at the edges seemed appropriate partly to match `Sprite`'s behavior, also it fit my usecase. 

A `TilemapAnchor` with `Anchor`'s variants and `None` or `TileCenter(u16,u16)` might be desirable then the it could default to `None` or `TileCenter(0,0)`, which would preserve the original behavior. That would clean up a little awkwardness when dealing with `Option<&Anchor>` in the code too since it could be part of the bundle.

However, one could also achieve a similar effect with a helper function that that produced an `Anchor::Custom(v)`: `get_tilemap_anchor(..) -> Anchor`.

> Do you envision the examples and suggested user usage to change to use Anchor::Center?

Yes, but only for the examples that are currently centering their tilemaps, which are the majority. I'd also suggest deprecating `get_tilemap_center_transform()` and placing a note to refer the user to `Anchor::Center`.

Oh! One big consideration I haven't looked at is how this works with non-rectangular tilemaps.

shanecelis:
Well, I took my own suggestions from above and am now using a `TilemapAnchor`. It was definitely more involved supporting all the layouts but I think the code actually came out better for it. The heart of it in `src/anchor.rs` is much more legible now. 

I decided to make use of the AABB code to generalize the behavior for all map types. The AABB seemed slightly bigger than it needed to be. I halved the size of the border. It's possible that's incorrect in its other uses, but I _believe_ it's just more tightly bound now. Before it had a half-tile-size of buffer around it for my purposes. If that's incorrect in its other uses, please tell me.

I also went through all the examples and I ran them to compare with previous behavior and new, and I believe they are all working even more complicated ones like "mouse_to_tile". I did add an "Enter" key press to "hex_neighbors" to exercise the different anchors like the "anchor" example. Here's a movie of it.

movie of hex neighbors example

I think this is in good shape now.

ChristopherBiscardi:
>  I'd also suggest deprecating get_tilemap_center_transform() and placing a note to refer the user to Anchor::Center

yeah that seems like the right path forward. Would be nice to drop it.

shanecelis:
I think the typical user code that used to center the tilemap is an improvement. It's more flexible, doesn't require any knowledge about the map size, type, or tile size. 
```diff
-            transform: get_tilemap_center_transform(&map_size, &grid_size, &map_type, 0.0),
+            anchor: TilemapAnchor::Center,
```
However, the code that needs to ascertain the position of tiles is not an unmitigated improvement. Currently it looks like this:

```rust
let tile_center = tile_pos.center_in_world(grid_size, map_type).extend(1.0);
            let transform = *map_transform * Transform::from_translation(tile_center);
            let anchor_offset = anchor.from_map(map_size, grid_size, tile_size, map_type);
            let transform =
                *map_transform * anchor_offset * Transform::from_translation(tile_center);
```

The `anchor_offset` is my addition. Perhaps it could be improved by adding `TilemapAnchor` as an argument to `TilePos::center_in_world()`.

* * *

Perhaps this would be a new feature appropriate for discussion in a separate issue or PR, but as I was considering how one could get the position of tiles especially now that anchor does complicate it, I thought, what if `bevy_ecs_tilemap` populated the tiles with the correct information _if_ the tile entity contained a `Transform` component? I like that the current design does not require a `Transform` on tiles, and I would not suggest using that transform for rendering but only as a user affordance, so they could get the entity and add a child to achieve tile relative positioning. I think this would simplify things and bolster the "tile per entity" aspect of `bevy_ecs_tilemap`.

shanecelis:
To address the problem I created and cited above with ascertaining the position of tiles, I no longer require the user to manipulate anchor and its associated transform directly, I added the arguments `map_size` and `anchor` to `TilePos::center_in_world()`, trying to follow the ordering of arguments in similar calls.

```diff
-let tile_center = tile_pos.center_in_world(grid_size, map_type);
+let tile_center = tile_pos.center_in_world(map_size, grid_size, map_type, anchor);
let transform = *map_transform * Transform::from_translation(tile_center.extend(1.0));
```
Hopefully this improves the ergonomics and makes `center_in_world()` live up to its name better.

# Unused Optimization
Anchor now produces a `Vec2` instead of a `Transform` behind `pub(crate) fn as_offset(...)`. In "prepare.rs" I multiply by a `Transform::from_translation(anchor_offset)` but I considered merely adding the offset to the `translation` field directly, which would be faster and effective for many usecases; however, it would prohibit the user from rotating or scaling their map's transform. I left a note in the code mentioning this tradeoff.

shanecelis:
Just to summarize where this PR in hopes it will be reviewed and accepted, here is the overview of what this PR's changes:

* It does make breaking changes, namely the `center_in_world()` function has two more parameters `map_size` and `anchor`, so I'd suggest bumping bevy_ecs_tilemap to version 0.16.0 if accepted.
* It updates all the examples that used `get_tilemap_center_transform()` to use `TilemapAnchor::Center`.
* It deprecates `get_tilemap_center_transform()` with a message that recommends using `TilemapAnchor::Center`.
* It adds a new example called "anchor" that changes a map's anchor when `Space` is pressed.
* It modified "hex_neighbors" example to change its anchor when `Enter` is pressed.

I have gone through all the examples and I believe it's working in all cases. If anyone sees any discrepancies or bugs, please post them here.

src/anchor.rs
```diff
@@ -0,0 +1,66 @@
...
+    TopRight,
+    /// Custom anchor point. Top left is `(-0.5, 0.5)`, center is `(0.0, 0.0)`. The value will
+    /// be scaled with the tilemap size.
+    Custom(Vec2),
+    // TileCenter(TilePos),
+}
+
+impl TilemapAnchor {
+    /// Provides an offset from the center of the bottom-left tile with no
+    /// anchor to the center of the bottom-left tile with the given anchor.
```
adrien-bon, line 30: Maybe elaborate a bit on this one, I had to read it several time (or maybe that's a reading skill issue :))
As I understand it, this function will give us the translation / offset between the given `TilemapAnchor` and `TilemapAnchor::None`, which is the reference for rendering. Is that correct ?

src/helpers/projection.rs
```diff
@@ -4,14 +4,29 @@ use crate::helpers::square_grid::diamond::DiamondPos;
...
     /// The center is well defined for all [`TilemapType`]s.
-    pub fn center_in_world(&self, grid_size: &TilemapGridSize, map_type: &TilemapType) -> Vec2 {
+    pub fn center_in_world(
+        &self,
+        map_size: &TilemapSize,
+        grid_size: &TilemapGridSize,
+        map_type: &TilemapType,
+        anchor: &TilemapAnchor,
+    ) -> Vec2 {
+        let offset = anchor.as_offset(map_size, grid_size, None, map_type);
```
adrien-bon, line 21: It feels odd that we can set an `Option<TilemapTileSize>` in `TilemapAnchor::as_offset()` but assume that for `TilePos::center_in_world()` it is alway the same as `TilemapGridSize`.
I think we should allow the user to provide a `TilemapTileSize` in both functions and just pass along the `TilemapTileSize`.

src/helpers/projection.rs
```diff
@@ -57,34 +72,39 @@ impl TilePos {
         map_size: &TilemapSize,
         grid_size: &TilemapGridSize,
         map_type: &TilemapType,
+        anchor: &TilemapAnchor,
     ) -> Option<TilePos> {
+        let offset = anchor.as_offset(map_size, grid_size, None, map_type);
```
adrien-bon, line 77: Same remark as above regarding the `Option<TilemapTileSize>`

adrien-bon (commented):
Added some comments but it's mostly nitpcking.
Very nice feature overhall ! :)

src/anchor.rs
shanecelis: Wow, I had to read that one three times before I understood what past-me meant. Will try again.

src/helpers/projection.rs
shanecelis: I need to setup--or find (anyone know?)--an example where the grid and tile sizes are different. I believe that the only case where it matters is when the tile size is bigger than the grid size, but I need to confirm that.

shanecelis: I haven't yet setup the example I want. But assuming it only matters when tile sizes are bigger than the grid, here's my question, should we align to

a) the corners of the tilemap's grid? or

b) the tilemap + tile that goes off the grid? 

We can do either. I'd prefer not to do both, and if we choose a) then we can drop caring about tile_size altogether in the `as_offset()` and `center_in_world()` API (I think).

adrien-bon: Yeah you are right, we should either do one or the other and a) would probably make more sense.

src/helpers/projection.rs
shanecelis: Resolved. It is no longer optional.

shanecelis:
After playing around with an example that altered the grid size, it seemed best to propagate tile size through the API. It aligns with intuitions and looks correct on inspection. I've done that and I modified the hex_neighbors example to react to the TAB key to change the grid scale between 1, 0.5, and 1.5. Here's a video of it in action:

hex_neighbors

I also modified the example to show the three parameters in play and their key bindings more clearly. Although I worry I've overloaded the hex_neighbors example well past its original reason for being. Perhaps I should try to merge the "anchor" and "hex_neighbors" into just one "anchor" example and restore the original "hex_neighbors".

I believe these changes address all of @adrien-bon's feedback. Thank you again @adrien-bon for the review.

src/render/extract.rs
```diff
@@ -336,7 +342,7 @@ pub fn extract(
     let extracted_tilemaps: Vec<_> = extracted_tilemaps.drain().map(|(_, val)| val).collect();
 
     // Extracts tilemap textures.
-    for (render_entity, _, tile_size, tile_spacing, _, _, texture, _, _, _, _) in
+    for (render_entity, _, tile_size, tile_spacing, _, _, texture, _, _, _, _, _anchor_maybe) in
```
rparrett, line 345: This value is no longer an option

src/render/extract.rs
```diff
@@ -326,7 +330,9 @@ pub fn extract(
                         visibility: *data.8,
                         frustum_culling: *data.9,
                         render_settings: *data.10,
+                        // anchor: *data.11.unwrap_or(&Anchor::BottomLeft),
```
rparrett, line 333: Some more leftovers

src/render/extract.rs
shanecelis: Good catch. Updated to '_'.

src/anchor.rs
```diff
@@ -0,0 +1,66 @@
+use crate::{prelude::chunk_aabb, TilemapGridSize, TilemapSize, TilemapTileSize, TilemapType};
+use bevy::prelude::*;
+use std::borrow::Cow;
+
+/// How a tilemap is positioned relative to its [`Transform`]. It defaults to
+/// `TilemapAnchor::None` which is the center of the bottom-left tile.
```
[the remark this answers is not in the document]
shanecelis: It's almost tempting to rename `None` to `TileCenterBottomLeft` but I think I'll hold out for issue #598. Note added.

shanecelis:
```sh
$ commit -m "final-final-complete-patch-last-for-sure.doc"
```
I replaced the "anchor.rs" example with the modified "hex_neighbors.rs" example. I added the square and isometric map types so it covers all types. Last two commits only touch example code. Here is the new "anchor" example in action:

new "anchor" example

src/anchor.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub fn as_offset(
+        &self,
+        map_size: &TilemapSize,
+        grid_size: &TilemapGridSize,
+        tile_size: &TilemapTileSize,
+        map_type: &TilemapType,
+    ) -> Vec2 {
+        // let tile_size = tile_size
+        //     .map(Cow::Borrowed)
+        //     .unwrap_or_else(|| Cow::Owned(TilemapTileSize::new(grid_size.x, grid_size.y)));
```
rparrett, line 52:
```suggestion
```

examples/hex_neighbors_radius_chunks.rs
```diff
@@ -312,24 +345,47 @@ fn swap_map_type(
 #[derive(Component)]
 struct TileLabel(Entity);
 
+const DEFAULT_FONT_SIZE: f32 = 14.0;
+const BIGGER_FONT_SIZE: f32 = 16.0;
```
rparrett, line 349: nit: `DEFAULT_` feels *slightly* confusing due to the default on `TextFont` being `20`. But I would personally ditch the new "increased font size on hover" effect. There just isn't much room in the tiles to increase the size.

rparrett (approved):
I haven't reviewed *quite* as thoroughly as I'd like, but I'm hitting the approve button to mark my general approval of the concept and make sure I'm not blocking anything while I'm away for a bit.

src/anchor.rs
shanecelis: Thank you. Removed.

examples/hex_neighbors_radius_chunks.rs
shanecelis: Removed the `BIGGER_FONT_SIZE`.

ChristopherBiscardi:
I love the removal of `get_tilemap_center_transform` and the code seems to work as expected. My only concern is thinking about future upstreaming/tilemap implementation efforts. afaict, Mesh2d doesn't respond to anchor upstream so this creates more distance from upstream who seems to be pursuing a greenfield base tilemap rendering implementation.

that said, this works and is useful so merging :)

Issue #597, opened 2025-02-20
Deprecate bundles in favor of Required Components


Johardt:
As of Bevy 0.15, Bundles were deprecated in Bevy with [Required Components](https://bevyengine.org/news/bevy-0-15/#required-components) replacing them. This plugin should follow suit. This would mean replacing Bundles like this:
```rs
#[derive(Bundle, Debug, Default, Clone)]
pub struct StandardTilemapBundle {
    pub grid_size: TilemapGridSize,
    pub map_type: TilemapType,
    pub size: TilemapSize,
    pub spacing: TilemapSpacing,
    pub storage: TileStorage,
    pub texture: TilemapTexture,
    pub tile_size: TilemapTileSize,
    pub transform: Transform,
    pub global_transform: GlobalTransform,
    pub render_settings: TilemapRenderSettings,
    /// User indication of whether an entity is visible
    pub visibility: Visibility,
    /// Algorithmically-computed indication of whether an entity is visible and should be extracted
    /// for rendering
    pub inherited_visibility: InheritedVisibility,
    pub view_visibility: ViewVisibility,
    /// User indication of whether tilemap should be frustum culled.
    pub frustum_culling: FrustumCulling,
    pub sync: SyncToRenderWorld,
}
```
With something like this:
```rs
#[derive(Component, Debug, Default, Clone)]
#[require(
    TilemapGridSize, 
    TilemapType, 
    TilemapSize, 
    TilemapSpacing, 
    TileStorage, 
    TilemapTexture, 
    TilemapTileSize, 
    Transform, 
    GlobalTransform, 
    TilemapRenderSettings, 
    Visibility, 
    InheritedVisibility, 
    ViewVisibility, 
    FrustrumCulling, 
    SyncToRenderWorld)]
pub struct StandardTilemap;
```

tigerplush:
I would appreciate this very much. When adding a TilemapBundle, it overrides my Transform, which is inconvenient

patrickariel:
I've opened a PR for this, if you want to test it out: https://github.com/StarArawn/bevy_ecs_tilemap/pull/604

ChristopherBiscardi:
> As of Bevy 0.15, Bundles were deprecated in Bevy

[...]

PR #604, opened 2025-04-13
feature: migrate to required components

patrickariel:
Closes https://github.com/StarArawn/bevy_ecs_tilemap/issues/597

# Objective

- Required components replaces bundles, but the crate still contains many user-facing bundles.

## Solution

- Deprecate the following bundles:
  - `TileBundle`
  - `MaterialTilemapBundle`
  - `TilemapBundle`
  - `StandardTilemapBundle`

- Migrate all examples to required components.
- Deprecate `MaterialTilemapHandle` in favor of `TilemapMaterial`. This is a pure name change done for two reasons:
  - To conform to the idiomatic pattern used [upstream](https://github.com/bevyengine/bevy/blob/71a9142deda547a872da3cc2b6a6300bad6e86a6/examples/3d/3d_shapes.rs#L83-L93):
    ```rust
    commands.spawn((
        Mesh3d(shape),
        MeshMaterial3d(material),
    ));
    ```
  - The material component is now something that is explicitly inserted by the user. So, we need to make sure that it's not cumbersome to remember and type.

## Testing

I've ran all the tests on Linux (Wayland) and they all look fine to me. But I'm not 100% familiar with all the quirks and edge-cases of this crate. So if I've overlooked any potential regressions, do let me know 🙂

## Migration Guide

The bundles are now all deprecated. With the new version, you should instead compose together `Tile`, `Tilemap` and `TilemapMaterial` to achieve your desired behavior.

Before:

```rust
commands.spawn(TileBundle::default());

commands.spawn(MaterialTilemapBundle {
    material,
    ..default(),
});

commands.spawn(TilemapBundle::default());

commands.spawn(StandardTilemapBundle::default());
```

After:

```rust
commands.spawn(Tile);

commands.spawn((Tilemap, TilemapMaterial(material)));

commands.spawn((Tilemap, TilemapMaterial::standard()));

commands.spawn(Tilemap);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,7 @@ use bevy::{
 };
 
 #[cfg(feature = "render")]
-use render::material::MaterialTilemapHandle;
+use render::material::TilemapMaterial;
 
 use anchor::TilemapAnchor;
 use map::{
@@ -112,11 +112,22 @@ impl Default for FrustumCulling {
 }
 
 #[cfg(feature = "render")]
+#[expect(deprecated)]
+#[deprecated(
+    since = "0.16.0",
+    note = "Use the `Tilemap` and `TilemapMaterial::standard()` components instead. 
+        Inserting `Tilemap` will now also insert all necessary components automatically."
+)]
 pub type TilemapBundle = MaterialTilemapBundle<StandardTilemapMaterial>;
 
 #[cfg(feature = "render")]
 /// The default tilemap bundle. All of the components within are required.
 #[derive(Bundle, Debug, Default, Clone)]
+#[deprecated(
+    since = "0.16.0",
+    note = "Use the `Tilemap` and `TilemapMaterial<M>` components instead. 
+        Inserting `Tilemap` will now also insert all necessary components automatically."
+)]
 pub struct MaterialTilemapBundle<M: MaterialTilemap> {
     pub grid_size: TilemapGridSize,
     pub map_type: TilemapType,
@@ -136,14 +147,37 @@ pub struct MaterialTilemapBundle<M: MaterialTilemap> {
     pub view_visibility: ViewVisibility,
     /// User indication of whether tilemap should be frustum culled.
     pub frustum_culling: FrustumCulling,
-    pub material: MaterialTilemapHandle<M>,
+    pub material: TilemapMaterial<M>,
     pub sync: SyncToRenderWorld,
     pub anchor: TilemapAnchor,
 }
 
+#[derive(Component, Debug, Default, Clone)]
+#[require(
+    TilemapGridSize,
+    TilemapType,
+    TilemapSize,
+    TilemapSpacing,
+    TileStorage,
+    TilemapTexture,
+    TilemapTileSize,
+    Transform,
+    TilemapRenderSettings,
+    TilemapAnchor,
+    Visibility,
+    FrustumCulling,
+    SyncToRenderWorld
+)]
+pub struct Tilemap;
+
 #[cfg(not(feature = "render"))]
 /// The default tilemap bundle. All of the components within are required.
 #[derive(Bundle, Debug, Default, Clone)]
+#[deprecated(
+    since = "0.16.0",
+    note = "Use the `Tilemap` component instead. 
+        Inserting `Tilemap` will now also insert all necessary components automatically."
+)]
 pub struct StandardTilemapBundle {
     pub grid_size: TilemapGridSize,
     pub map_type: TilemapType,
@@ -169,8 +203,11 @@ pub struct StandardTilemapBundle {
 /// A module which exports commonly used dependencies.
 pub mod prelude {
     #[cfg(feature = "render")]
+    #[expect(deprecated)]
     pub use crate::MaterialTilemapBundle;
+    pub use crate::Tilemap;
     #[cfg(feature = "render")]
+    #[expect(deprecated)]
     pub use crate::TilemapBundle;
     pub use crate::TilemapPlugin;
     pub use crate::anchor::TilemapAnchor;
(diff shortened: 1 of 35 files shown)


src/lib.rs
```diff
@@ -112,11 +112,22 @@ impl Default for FrustumCulling {
...
 #[cfg(feature = "render")]
+#[expect(deprecated)]
+#[deprecated(
+    since = "0.16.0",
+    note = "Use the `Tilemap` and `TilemapMaterial::standard()` components instead. 
+        Inserting `Tilemap` will now also insert all necessary components automatically."
+)]
 pub type TilemapBundle = MaterialTilemapBundle<StandardTilemapMaterial>;
 
 #[cfg(feature = "render")]
```
ConnerPetzold, line 123: I think to maintain feature parity, you'd need to create two versions of the `Tilemap` component based on the render feature.

src/lib.rs
```diff
@@ -136,14 +147,37 @@ pub struct MaterialTilemapBundle<M: MaterialTilemap> {
...
+    TilemapSpacing,
+    TileStorage,
+    TilemapTexture,
+    TilemapTileSize,
+    Transform,
+    TilemapRenderSettings,
+    TilemapAnchor,
+    Visibility,
+    FrustumCulling,
+    SyncToRenderWorld
```
ConnerPetzold, line 169: I wonder if `Tilemap` should require `TilemapMaterial<StandardTilemapMaterial>`? I'm not sure how generics work with required components but it seems like the vast majority of users won't need a custom material.

src/lib.rs
patrickariel: I believe the feature split shouldn't be necessary anymore. `StandardTilemapBundle` (`not(feature = "render")`)  was just `MaterialTilemapBundle` (`feature = "render"`) without `TilemapMaterial`.

src/lib.rs
patrickariel: I think this would sort of complicate the API surface, because now we would need a `TilemapWithoutMaterial` for users who wants custom material or no material (e.g. `not(feature = "render")`).

Having the material component be opt-in feels a lot more natural and idiomatic, in line with how it is handled upstream:

```rust
commands.spawn((
    Mesh3d(shape),
    MeshMaterial3d(material),
));
```

Here, `Mesh3d` does not automatically require `MeshMaterial3d<M>`. It must be inserted manually by the user.

ConnerPetzold: Hmm yeah I was only thinking about the use case of a custom material and not about the non-rendering case. I suppose that case will be more common for headless/backend.

ConnerPetzold: Although why not just make `Tilemap` conditionally require `TilemapMaterial<StandardTilemapMaterial>` based on the render flag to mirror what happens now? That way you don't need two separate components. And I think it still covers the use case of a custom material because I believe you'd still be able to override with a custom `TilemapMaterial`, though I'm not 100% sure about that.

teohhanhui: > I believe you'd still be able to override with a custom `TilemapMaterial`

That's correct.

patrickariel: > I believe you'd still be able to override with a custom `TilemapMaterial`, though I'm not 100% sure about that.

Unfortunately, this will not work. `TilemapMaterial<A>` and `TilemapMaterial<B>` are different types and will be treated as different components.

If you have a component like this:

```rust
#[derive(Component)]
#[require(TilemapMaterial<StandardTilemapMaterial>)]
struct Tilemap;
```

And you insert it like this:

```rust
commands.spawn((Tilemap, TileMapMaterial::<Custom>::default()));
```

You will end up with an entity that looks like this:

```rust
(
    Tilemap,
    TileMapMaterial<Custom>,
    TilemapMaterial<StandardTilemapMaterial>,
)
```

patrickariel: > We should follow the same pattern, i.e. the user needs to spawn `TilemapMaterial` by themselves when spawning `Tilemap`?

Yep, that's what I'm currently proposing.

ConnerPetzold: Ok here's my proposal:

- `Tilemap` conditionally requires `TilemapTexture` based on the `render` feature
- Make `TilemapMaterial` an optional component and remove `TilemapMaterial<StandardTilemapMaterial>` entirely
- If the entity has a `TilemapMaterial` component, use that as the material, otherwise fall back to using the `StandardTilemapMaterial` resource

I think that would allow:

```rust
(Tilemap) // Use the standard material
(Tilemap, TilemapMaterial<Custom>) // Use a custom material
```

ConnerPetzold: Ah yeah I see, that's a tricky problem..  What about something like this:
```rust
#[cfg(feature = "render")]
#[derive(Component, Debug, Default, Clone)]
#[require(
    TilemapGridSize,
    TilemapType,
    TilemapSize,
    TilemapSpacing,
    TileStorage,
    TilemapTexture,
    TilemapTileSize,
    Transform,
    TilemapRenderSettings,
    TilemapAnchor,
    Visibility,
    FrustumCulling,
    SyncToRenderWorld
)]
pub struct TilemapMaterial<M: MaterialTilemap>(pub Handle<M>);

#[cfg(feature = "render")]
pub type Tilemap = MaterialTilemapBundle<StandardTilemapMaterial>;

#[cfg(not(feature = "render"))]
#[derive(Component, Debug, Default, Clone)]
#[require(
    TilemapGridSize,
    TilemapType,
    TilemapSize,
    TilemapSpacing,
    TileStorage,
    TilemapTileSize,
    Transform,
    Visibility,
)]
pub struct Tilemap;
```

teohhanhui: @ConnerPetzold That just doesn't make sense and is super confusing for users. Why not just stick to the same pattern as Bevy? As a user I'd like better DX too, but I think this should be done at the Bevy level.

ConnerPetzold: I don't understand why it's confusing for users:
```rust
(TIlemap) // use the standard material if render=true, otherwise no materials/rendering
(TilemapMaterial<Custom>) // use a custom material, and will only be available if render=true
```

It's mirroring what ecs_tilemap has now, and essentially just collapses what you have with `Tilemap` and `TilemapMaterial`. I don't think the `MeshMaterial3d` pattern applies here because writing a custom tilemap material will be very rare. I think the more appropriate Bevy pattern to follow here would be `Sprite`, since like tilemaps you almost never need a custom material (and I guess in the case of Sprite, you can't even use a custom material). For the rare cases you do, you have the mesh/material escape hatch.

teohhanhui: I guess that would make sense if it's called `TilemapWithMaterial`.

suprohub:
any updates ?

PR #605, opened 2025-04-14
Update to bevy 0.16rc4

EmiOnGit:
The only breaking change I found is the renaming of the `bevy_log` feature to `tracing` from this [pr](https://github.com/bevyengine/bevy/commit/c5365509bfbef5f19345b84b51258ff37944fe72).

I did test the majority of examples on linux

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,12 +16,12 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.16.0-rc.3", default-features = false, features = [
+bevy = { version = "0.16.0-rc.4", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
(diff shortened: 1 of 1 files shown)


ChristopherBiscardi:
Thank you for the PR!

Unfortunately the PR that this addresses was reverted in https://github.com/bevyengine/bevy/pull/18816 soon after rc.4 went out

As such I generally consider rc.4 to be a buggy release when it comes to `bevy_log` and the implemented tracing/log functionality specifically and don't think we should do a release that we'll have to revert fairly immediately. I'll leave this open until rc.5 (or stable) is released so we don't forget to revisit and check this behavior in the next candidate.

musjj:
`rc.5` just landed :grin:

Jarten0:
would #607 close this then?

ChristopherBiscardi:
yes, I've merged #607

PR #606, opened 2025-04-15
chore: bump Bevy to v0.16.0-rc4 and fix collection imports

miketwenty1:


.DS_Store
rparrett, line 1: Stray file

miketwenty1: thanks! also im going to close this because im an idiot didn't see we already had a PR for rc4. I'm going to reopen one for rc5

PR #608, opened 2025-04-15
bump to rc.2 for bevy rc.5 support

ChristopherBiscardi:
doing a release, so merging the version tag


rparrett (approved):
Spot checked a few examples on web and native macos

PR #610, opened 2025-04-24
chore: bump bevy to 0.16.0 release tag

miketwenty1:
Bevy 16.0 release version bump.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ render = []
 serde = ["dep:serde"]
 
 [dependencies]
-bevy = { version = "0.16.0-rc.5", default-features = false, features = [
+bevy = { version = "0.16.0", default-features = false, features = [
     "bevy_core_pipeline",
     "bevy_render",
     "bevy_asset",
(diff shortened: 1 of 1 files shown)


jfouche:
Shouldn't you use "0.16" instead of "0.16.0" in case of bevy 0.16 patches, as 0.16.x bevy patches shouldn't break API ?

rparrett:
> Shouldn't you use "0.16" instead of "0.16.0" in case of bevy 0.16 patches, as 0.16.x bevy patches shouldn't break API ?

They are equivalent. See [The Cargo Book](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio).

miketwenty1:
fwiw, @jfouche I also didn't know this. So, the patch number is "this or better" for the patch range. 0.16.0 is all the way up to <0.17.0. if we instead specified 0.16.1 this would not include 0.16.0 as a valid dep, but would now include 0.16.1 to <0.17.0. So like rparrett says they are equivalent because 0.16 would equate to 0.16.0, meaning the whole patch range.

If we wanted to peg specifically to a patch version it would look like this `"=0.16.0"`.
if it's preferred to have `0.16` as cleaner version syntax let me know, I'd be happy to adjust the PR.

ChristopherBiscardi (approved):
thanks!

PR #612, opened 2025-05-06
fix(#611): enable Bevy `serialize` feature when building with `serde` feature

adrien-bon:
Fix #611

Just enable the `serialize` feature from Bevy when we build bevy_ecs_tilemap with the `serde` feature.

PR #613, opened 2025-05-06
bump to v16

ChristopherBiscardi:
The main difference from the release candidates is that this release includes #612


rparrett (approved):
Still looking good on my end.

PR #617, opened 2025-06-21
Wrote test suite; ran rustfmt

axle12693:
Wrote the beginnings of a test suite - not comprehensive, but I hope it helps with dev speed in the future.

cargo test --lib --tests


axle12693:
Apologies for having to add multiple fixes.

At this point, I think the tests are failing because the machine doesn't have enough memory to compile everything with debug.

What action should be taken going forward?

PR #623, opened 2025-09-13
0.17 update

ChristopherBiscardi:
released on crates.io as rc.1

### DIFF
--- a/src/map.rs
+++ b/src/map.rs
@@ -193,19 +193,16 @@ impl TilemapTexture {
         for handle in self.image_handles() {
             // NOTE: We retrieve it non-mutably first to avoid triggering an `AssetEvent::Modified`
             // if we didn't actually need to modify it
-            if let Some(image) = images.get(handle) {
-                if !image
+            if let Some(image) = images.get(handle)
+                && !image
                     .texture_descriptor
                     .usage
                     .contains(TextureUsages::COPY_SRC)
-                {
-                    if let Some(image) = images.get_mut(handle) {
+                    && let Some(image) = images.get_mut(handle) {
                         image.texture_descriptor.usage = TextureUsages::TEXTURE_BINDING
                             | TextureUsages::COPY_SRC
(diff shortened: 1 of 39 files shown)


src/map.rs
```diff
@@ -195,32 +193,28 @@ impl TilemapTexture {
...
-                TilemapTexture::Vector(handles.iter().map(|h| h.clone_weak()).collect())
-            }
+            TilemapTexture::Vector(handles) => TilemapTexture::Vector(handles.to_vec()),
             #[cfg(not(feature = "atlas"))]
             TilemapTexture::TextureContainer(handle) => {
-                TilemapTexture::TextureContainer(handle.clone_weak())
+                TilemapTexture::TextureContainer(handle.clone())
             }
         }
     }
```
rparrett: Is this function still useful after this change? At the very least, its name is now not ideal.

bbarker:
@ChristopherBiscardi I was taking a look at updating the release version of 0.17: https://github.com/ChristopherBiscardi/bevy_ecs_tilemap/pull/3

ChristopherBiscardi:
@bbarker that PR has so many formatting changes its hard to understand what is being changed. I'm not aware of any functional changes being needed at the moment for 0.17's release, is there something in that PR that needs to be merged besides removing the -rc postfix?

ChristopherBiscardi:
There was a kind of bizarre clippy issue that registered items as dead code even though they were used in the examples helpers. I think this was due to the code organization of the helpers directory, but I'm leaving a refactor to remove them to after release.

src/map.rs
ChristopherBiscardi: I've removed it.

ChristopherBiscardi:
github actions is being really flakey, but I don't believe there are actual issues since similar jobs are passing on other platforms. Notably I had to re-run the windows job, and the linux jobs are failing with a linker error.

rparrett (approved):

PR #629, opened 2026-01-15
Bevy 0.18

agluszak:
Done by @ChristopherBiscardi, I only bumped the version from rc to release :)


ChristopherBiscardi:
thank you! but I'm going to close this since we don't really need two prs with all of the changes, and the other one is the one we're doing 0.18 work on.

